PR #21, opened 2015-09-25
VOMS Handling

sustr4:
Improved handling in the VOMS slave:

- Overall return value
- Processing exceptions form voms-admin
- More checks

Keep open for now, but comment if you want.

### DIFF
--- a/slave/process-voms.sh
+++ b/slave/process-voms.sh
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+# Script for managing user membership in VOs
+# Return values:
+# 0 ... OK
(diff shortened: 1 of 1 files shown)


sustr4:
IMO this pull request is now ready to be merged. Changes made:

- All calls to voms-admin checked for return value
- All loops rewritten to avoid running in subshells, so that return values can propagate back to the calling function
- All routine reporting redirected to STDERR
- Error logging through syslog as well as Perun's log_msg()
- Hierarchical return value returned -- the higher the number, the bigger the issue. We can consider changing that to a bit-wise flag, though.

PR #22, opened 2015-09-30
Remove prefix of userCertDNs subjectDNs for all services where needed

stavamichal:
- remove prefix before writing to the file
 - in metacloud export change uniqueness only if two different users has
   the same subjectDN (not if the same user has two or more same
   subejctDNs)


stavamichal:
Do not merge without merging also changes in core in: https://github.com/CESNET/perun/pull/987

PR #23, opened 2015-09-30
Set also ownership when use diff_mv

stavamichal:
- when use function diff_mv, set also ownership, not just permissions
   of destination file to source file


zlamalp:
This change was merged locally by Slávek a week ago. Closing this pull-request.

PR #35, opened 2015-11-23
Local settings to skip certain VOs

sustr4:
- Skip deleting/creating members for certain VOs
- Variables introduced to skip whole VO creation/deletion [NOT YET IMPLEMENTED]


sustr4:
This is ready to merge!

slave/process-voms.sh
```diff
@@ -42,6 +42,22 @@ function process {
 
 	# Iterate through every VO and check which user will be added or removed
 	for VO in `echo -e $VOS`; do
+		# First, check if some part should be skipped
+		if echo ${NO_CREATE_MEMBERS} | grep -wq ${VO}; then
```
licehammer: grep's -w switch uses also '.' as a word delimiter. Therefore when I want to skip VO "test.myvo.cz" then "myvo.cz" will be also skipped.

sustr4: Ah, didn't test for that. Will fix.

sustr4:
OK
- fixed token detection
- added detection and better handling of empty VOs
Yours again :wink:

slave/process-voms.sh
```diff
@@ -42,17 +42,44 @@ function process {
 
 	# Iterate through every VO and check which user will be added or removed
 	for VO in `echo -e $VOS`; do
+		# First, check if some part should be skipped
+		if echo ${NO_CREATE_MEMBERS} | grep -E "([[:space:]]|^)${VO}([[:space:]]|$)"; then
```
licehammer: This can still causes trouble if $VO contains any character which will be interpreted as REGEXP special char, e.g. dot.

Something like this could work:
```bash
echo ${NO_CREATE_MEMBERS} | grep -E '([[:space:]]|^)'` echo ${VO} | sed -e 's/\./\\./g' `'([[:space:]]|$)'
```

or something like this:
```bash
echo "$VO" | grep -x -F "$NO_CREATE_MEMBERS"
```
but it requires that items in $NO_CREATE_MEMBERS will be separated by newline

sustr4:
Sorry, I know this is becoming repetitive. Now, though, special characters should be treated right in the needle as well as the haystack.

sustr4:
We have decided not to introduce this functionality after all. A more elaborate pull request will arrive in a moment with the whole VOMS connector reimplemented.

PR #42, opened 2016-01-07
Use also the old path /opt/perun/bin/ for pre/mid/post scripts

stavamichal:
Some changes about using /opt/perun/bin/ for pre/mid/post scripts
    
     - use scripts from both paths /opt/perun/bin/{service}.d/ and
       /etc/perun/{service}.d/
     - if two scripts has the same name, higher priority has new path script
     - change path to lib to /opt/perun/lib/{service}/ from
       'process-{service}'
     - Fix Makefile and generator of RPM to use right path (wrong path is
       /opt/perun/bin/process-{service}.d/ and
       /opt/perun/lib/process-{service}/
     - add info to all affected changelogs
     - fix ldap and ldap-svb-vi scripts to work with new correct paths

### DIFF
--- a/slave-new/Makefile
+++ b/slave-new/Makefile
@@ -29,7 +29,7 @@ meta:
 
 $(processes):
 	@echo "Generating deb package for $@..."
-	@cd $@ && NAME=$@ envsubst < ../templates/Makefile > Makefile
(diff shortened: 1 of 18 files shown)


licehammer:
merged locally

PR #55, opened 2016-02-25
VOMS reimplement

sustr4:
The central functionality seems to be in place. I need assistance integrating with perun services, namely:

* how to call the perl slave script
* how to distribute its dependences


stavamichal:
Closing this request, because fixed version of this code is in new request #70

PR #56, opened 2016-03-08
New generic gen for generating data to binary file

stavamichal:
- use option "-s" to specify service name
- store data to /tmp/"service"-data file
- these data can be used by some gen using option "-d" and "-F"

### DIFF
--- a/gen/generic_gen
+++ b/gen/generic_gen
@@ -1,0 +1,17 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use perunServicesInit;
+use Storable qw(store_fd);
+
+local $::PROTOCOL_VERSION = "3.0.0";
+
+perunServicesInit::init;
+
+my $DIRECTORY = perunServicesInit::getDirectory;
+my $data = &$::GET_DATA_METHOD;
+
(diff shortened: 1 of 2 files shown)


gen/generic_gen
```diff
@@ -0,0 +1,17 @@
...
+use strict;
+use warnings;
+use perunServicesInit;
+use Storable;
+
+local $::PROTOCOL_VERSION = "3.0.0";
+
+perunServicesInit::init;
+my $DIRECTORY = perunServicesInit::getDirectory;
+my $data = perunServicesInit::getHierarchicalData;
```
licehammer: This have to be configurable. For example you might need getDataWithGroups

gen/generic_gen
```diff
@@ -0,0 +1,17 @@
...
+
+my $file = "/tmp/$::SERVICE_NAME-data";
```
licehammer: I would prefer to chose file name or just print it to stdout and let user to decide where he wants it.

stavamichal:
Behavior was changed based on comments. @licehammer  can you check it again?

Thank you.

PR #71, opened 2016-04-07
New version of voms gen script

stavamichal:
- using groups and roles
 - using new attributes: vomsRoles (G,R), vomsGroupName (G)
 - generate xml structure instead of list of certificates with mails
 - change protocol version


stavamichal:
Do not merge without consultation with Michal S. or Slavek L. Waiting for new slave script voms.

PR #77, opened 2016-04-25
Fixed services build using dh_make on Ubuntu 16.04+

zlamalp:
- With newer version of dh_make we can't use -i and -s options together.
  Each of them states to produce a single deb package.
  I kept "-i" since it means "single arch-independent package", while "-s" stated
  "It is the standard case, so if you don't know what to do, choose this".
- Not tested on older systems, but result should be the same.
  We want arch-independent package and it can be now only of 1 class.


licehammer:
Please fix typo in commit message:
With newer version of dh_make we can't use**r**

Except of this, it's good to go.

zlamalp:
Commit message fixed.

PR #79, opened 2016-05-02
Add better timeout and test all hosts for perun_propagate

stavamichal:
- timeout doesn't end propagation, propagation will continue in
       background, only perun_propagation ends with info about it
     - try to remctl for every uniqe host in ktutils because of more hosts
       for one physical address

### DIFF
--- a/other/perun-propagate/etc/perun_propagate
+++ b/other/perun-propagate/etc/perun_propagate
@@ -12,12 +12,15 @@
 # Description:       Start the perun propagate process
 ### END INIT INFO
 
+KEYTAB=/etc/krb5.keytab
 NAME=perun-propagate
 DESC="Perun propagation"
 PERUN_SERVER='perun.ics.muni.cz'
+#Timeout in seconds
 TIMEOUT=120
 LOG=/var/log/$NAME
-KEYTAB=/etc/krb5.keytab
+#Remove content of old log file
+> "$LOG"
 
 umask 077
 
@@ -29,6 +32,12 @@ fi
 # Load lsb functions
 . /lib/lsb/init-functions
 
+#---------------------------- ADDITIONAL FUNCTIONS ----------
+# Return 0 if process with PID is still runing, 1 if not
+is_process_running() {
+	kill -0 "$1" 2>/dev/null 1>&2 && return 0 || return 1
+}
+
 check() {
 	for C in ktutil kinit remctl; do
 		which "${C}" &>/dev/null || {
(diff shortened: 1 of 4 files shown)


other/perun-propagate/etc/perun_propagate
```diff
@@ -29,6 +29,16 @@ fi
 # Load lsb functions
 . /lib/lsb/init-functions
 
+# Return 1 if process with PID is still runing, 0 if not
```
licehammer: Usually return 0 means everything is OK, return non-zero means that something is wrong or at least it isn't as you expect (e.g. process is not running).

other/perun-propagate/etc/perun_propagate
```diff
@@ -55,12 +65,26 @@ perun_propagate_start() {
...
+
+	if [ -z "${EXIT_CODE}" ]; then
+		OUT="Timeout elapsed, but script is still running."
+		echo "$OUT" >&2
+		EXIT_CODE=1;
 	else
-		OUT=`remctl "${PERUN_SERVER}" perun propagate 2>&1`
-		EXIT_CODE=$?
+		OUT=`cat "$TEMP_MSG"`
+		rm -f "$TEMP_MSG"
```
tauceti2: Don't know what this piece of code is really doing. Also please add restart to the start section.

stavamichal:
Fixed. Can you check please if it is correct now?

stavamichal:
New behavior for Principals added. Now all Principals from ktutil list will be tried.

stavamichal:
Added new version of this pull request. 
Can you check it again please?

other/perun-propagate/etc/perun_propagate
```diff
@@ -43,37 +56,111 @@ perun_propagate_start() {
...
+			SCRIPT_EXIT_CODE=$?
+			if [ "$SCRIPT_EXIT_CODE" = 0 ]; then
+				#scripts ends without error, so just add info from TMP FILE and set exit code to 0 if not in error yet
+				OUT="Perun_propagate for host ${PRINCIPAL_ARRAY[$i]} ends with status 0.\n"
+				if [ "${EXIT_CODE}" != 1 ]; then
+					EXIT_CODE=0
+				fi
+			else
+				#script ends with error so we need to get info if this error is just a problem with host or some other problem
+				grep -q "doesn't have assigned any facility" "${TMP_ERROR_ARRAY[$i]}"
```
tauceti2: If it cannot be done differently, I propose to set the text "doesn't have assigned any facility" as a constant at the beginning of the script. In case of change of the error message it will be little bit easier to find out why the condition isn't working.

other/perun-propagate/etc/perun_propagate
```diff
@@ -29,6 +32,16 @@ fi
...
 . /lib/lsb/init-functions
 
+# Return 0 if process with PID is still runing, 1 if not
+is_process_running() {
+	kill -0 "$1" 2>/dev/null 1>&2
+	if [ "$?" -eq 0 ]; then
+		return 0;
+	else
+		return 1;
+	fi
```
licehammer: Could be written in more compact way.
return $? # may return other values then 1 for false, but it usually doesn't matter.

or

kill  -0 "$1" 2>/dev/null 1>&2 && return 0 || return 1

other/perun-propagate/etc/perun_propagate
```diff
@@ -39,41 +52,116 @@ check() {
...
-		EXIT_CODE=$?
-	else
-		OUT=`remctl "${PERUN_SERVER}" perun propagate 2>&1`
-		EXIT_CODE=$?
+	ITERATOR=0;
+	while read -r LINE; do
+		if [ "${KERBEROS_TYPE}" = "heimdal" ]; then
+			kinit -k -t "${KEYTAB}" "${LINE}"
+		else	
+			kinit --no-afslog --keytab="${KEYTAB}" "${LINE}"
```
licehammer: Each kinit have to use separate file to store ticket (KRB5CCNAME). Otherwise is possible that the ticket will be overridden before remctl (spawned as separate process) manage to read it.

other/perun-propagate/etc/perun_propagate
```diff
@@ -39,41 +52,116 @@ check() {
...
+		fi
+		TMP_ERROR=`mktemp perun-propagate.XXXXXXXXX`
+		remctl "${PERUN_SERVER}" perun propagate 2>&1 | tee ${TMP_ERROR} >> ${LOG} &
+		PID=$!
+		PID_ARRAY[$ITERATOR]=$PID
+		TMP_ERROR_ARRAY[$ITERATOR]=$TMP_ERROR
+		PRINCIPAL_ARRAY[$ITERATOR]=$LINE
```
licehammer: I cannot find ITERATOR increment anywhere.

other/perun-propagate/etc/perun_propagate
```diff
@@ -39,41 +52,116 @@ check() {
...
+	#Prepare firt run of while with same START_TIME
+	NOW_TIME=$START_TIME
+	#Count of all processes in PIDS_ARRAY to know how much of them still running
+	PROCESSES_COUNT=${#PID_ARRAY[@]}
+	#Waiting for all running processes but not more than TIMEOUT in seconds
+	while [ `expr $NOW_TIME - $START_TIME` -lt "$TIMEOUT" ]; do
+		DONE_PROCESSES=0
+		for i in ${!PID_ARRAY[*]}; do
+			if is_process_running ${PID_ARRAY[$i]} ; then
+				break
```
licehammer: Consider adding short sleep before this "break". This script don't have to consume 100% processor time.

other/perun-propagate/etc/perun_propagate
stavamichal: Good point!

stavamichal:
Changed and fixed, please check now.

other/perun-propagate/etc/perun_propagate
```diff
@@ -38,42 +47,134 @@ check() {
...
+		if [ "$DONE_PROCESSES" -eq "$PROCESSES_COUNT" ]; then
+			break
+		fi
+		NOW_TIME=`date +%s`
+	done
+	#Get all statuses and errors and prepare messages and exit code
+	for i in ${!PID_ARRAY[*]}; do
+		is_process_running ${PID_ARRAY[$i]}
+		#process $i is still running
+		if [ "$?" = 0 ]; then
```
licehammer: This should work, but for numeric comparison is better to use "eq" than "=".

other/perun-propagate/etc/perun_propagate
```diff
@@ -38,42 +47,134 @@ check() {
...
+		fi
+		NOW_TIME=`date +%s`
+	done
+	#Get all statuses and errors and prepare messages and exit code
+	for i in ${!PID_ARRAY[*]}; do
+		is_process_running ${PID_ARRAY[$i]}
+		#process $i is still running
+		if [ "$?" -eq 0 ]; then
+			#if still running, we want to belive this is ok so we set exit code to 0 if not in error yet
+			OUT="Perun_propagate for host ${PRINCIPAL_ARRAY[$i]} still running even after timeout.\n"
```
licehammer: Don't you want to append to OUT instead of rewriting it?

stavamichal: Very good point. Fixed.

stavamichal:
Done.

PR #102, opened 2016-08-10
Fix changelogs for automatic generation of config directory

stavamichal:
- config directory is the one in /etc/perun/XY.d


licehammer:
Merged as part of dee72e1bf4979c5bc201a36e34c5fa22165cd4ee

PR #109, opened 2016-09-06
do not create afs volume when empty afs attributes

mvocu:
skip afs volume creation when appropriate AFS_ parameters are empty


stavamichal:
Please rebase and add missing changelog for this pull request.

Thank you.

mvocu:
Rebased.

Can you clarify the changelog thing?

stavamichal:
@mvocu 
you need to increase a version in slave/process-afs/changelog
only this way we can update this change to the package

mvocu:
Changelog modified.

zlamalp:
@licehammer @mvocu Maybe we could also update gen script, because right now it will print warnings when empty/undef value is printed, as allowed by new design.

zlamalp:
We should consult AFS expert.

PR #110, opened 2016-09-07
added information about updated version of script in git kypo2

frhrdina:
The script to import data from perun was uploaded to git of project kypo2, where the actual version of script will be kept.


licehammer:
@frhrdina: What about we delete this file entirely from our repo? Could you provide link to kypo2 git repo? I would put it into commit message.

frhrdina:
@licehammer I discussed it with Michal P. and we decided to delete this file. I can't provide link to kypo2 git repo because it is not public.

licehammer:
@frhrdina Do you want to delete only this file or whole slave script for KYPO portal? I'm wondering if it makes sense to create installation package without this file. What do you think?

licehammer:
After discussions with @frhrdina we decided that we keep kypo-portal service in this repo.

PR #112, opened 2016-09-14
Fix problem with 'NULL' in some core attributes

stavamichal:
- especialy in firstName and lastName (in perun these attributes are
  not mandatory)
- behavior is service specific
  - for some services return empty string ''
  - for some services return string 'null'
  - for other just skip empty value (do not add to file)

### DIFF
--- a/gen/appDB
+++ b/gen/appDB
@@ -79,10 +79,14 @@ for my $voData (@vosData) {
 				          };
 			}
 
+			my $userFirstName = defined($memberAttributes{$A_USER_FIRST_NAME}) ? $memberAttributes{$A_USER_FIRST_NAME} : '';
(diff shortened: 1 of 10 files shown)


zlamalp:
Hi @licehammer, for *ldap, crm_ceitec and ldap_ad_ceitec* are changes OK. For ldap, ``sn`` is required attribute based on name, if empty, it's correct that script will fail (nobody uses ldap service anyway - it's mostly template). As for crm_ceitec, sending ``'null'`` restores original behavior of an attribute. For ldap_ad_ceitec, ``displayName`` is optional param and if empty, it's not present in ldif. For ceitec service, name is always present in user, since based on that their CN/login is generated.

stavamichal:
Closing, because it is not actual against repo. Need to be create again.

PR #118, opened 2016-09-30
Gen scripts switched to user:virt:kerberosLogins

balcirakpeter:
Attribute urn:perun:user:attribute-def:def:kerberosLogins switched to virt
Change made in gen scripts: group_nfs4, k5login, passwd_nfs4, pkinit


balcirakpeter:
Related to pull request in CESNET/perun/Method for user:virt:kerberosLogins

PR #129, opened 2016-10-26
Cleanup - removed deprecated package sources

zlamalp:
- Removed /perun-propagate deb package sources.
- Last change was done to /other/perun-propagate instead.
- Removed old slave scripts in /slave.
- /slave-new moved to /slave.
- Updated readme.
- Removed deprecated perun-slave-meta package sources.


zlamalp:
For now, we can't remove ``slave-meta`` sources. So we can merge this as it is.

PR #139, opened 2016-11-08
Modification in passwd_nfs4 gen scripts

stavamichal:
- same changes for passwd_nfs4 and passwd_nfs4_mu
 - use defined attribute kerberos logins, because in virtual attribute
   there can be some other not supported logins for this service
 - change behavior of filtering duplicits in logins and creating logins
   with prefix 'nfs/'. Now we use a map with all kerberos logins to
   avoid duplicits and also create a 'nfs/login' equivalent for every
   'login' there


stavamichal:
Fixed. It was probably typo. Can you merge it now please?

PR #140, opened 2016-11-08
Add already existing service cloudIDP to GIT repository

stavamichal:
- gen script and send script

### DIFF
--- a/gen/cloudidp
+++ b/gen/cloudidp
@@ -1,0 +1,58 @@
+#!/usr/bin/perl
+use strict;
+use warnings;
+use perunServicesInit;
+use perunServicesUtils;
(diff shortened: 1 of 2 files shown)


licehammer:
Merged locally with indentation and whitespaces fixes.

zlamalp:
Hi @stavamichal and @licehammer, I see I'm late for the party, but this service should be removed. We do not have "slave" part of this. It was planned to be used on own cloudIdP instance, never happened. It's just slightly extended stub of generic ldap service. I believe we can also remove related facility and vo in CESNET instance of Perun (no destination -> ldap VM doesn't exists anymore). Not connected to external users (admins are @licehammer and @tauceti2).

PR #155, opened 2016-11-21
Use more compatible variant of using '~~' without warnings

stavamichal:
- '~~' is smartmatch in perl
 - we need to have compatible variant of using it between more versions
   of perl
 - this one seems to be most compatible
 - old one not work under perl version lower than 5.18

### DIFF
--- a/gen/fs_home
+++ b/gen/fs_home
@@ -1,10 +1,10 @@
 #!/usr/bin/perl
 use strict;
 use warnings;
-use experimental 'smartmatch';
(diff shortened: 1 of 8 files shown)


zlamalp:
Exactly the solution I found :-). Can you please also update files in ``send/`` folder for ldap and AD? They currently have older version of command, which works on idm instance:
```perl
no if $] >= 5.018, warnings => "smartmatch";
```
Thank you.

PR #160, opened 2016-12-01
Fixed working with groups and subgroups

frhrdina:
Fixed working with groups and subgroups, names of subgroups are imported into DB without prefix of parent groups


frhrdina:
Hi, OK. I updated it.

PR #168, opened 2017-01-05
New service open_nebula

stavamichal:
- this service is new version of metacloud and fedcloud export services
   with more specific behavior and JSON output

### DIFF
--- a/gen/opennebula
+++ b/gen/opennebula
@@ -1,0 +1,98 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use perunServicesInit;
+use perunServicesUtils;
+use Perun::Agent;
+use JSON::XS;
+
+our $SERVICE_NAME = "opennebula";
+our $PROTOCOL_VERSION = "3.0.0";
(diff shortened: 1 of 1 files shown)


gen/opennebula
```diff
@@ -0,0 +1,131 @@
...
+foreach my $resourceData ($data->getChildElements) {
+	processResourceData $resourceData;
+}
+
+my @users = values %$dataByUser;
+my %outStruc = ( "users" => \@users, );
+
+my $fileName = "$DIRECTORY/$SERVICE_NAME";
+open FILE,">$fileName" or die "Cannot open $fileName: $! \n";
+print FILE JSON::XS->new->utf8->pretty->encode(\%outStruc);
```
zlamalp: You can get json ouput sorted by keys through whole structure using ``...->pretty->canonical->encode(...``. It helps with future debug of changes in generated data.

stavamichal: Good point, thanks.

stavamichal:
Fixed.

Also added:
- do not use camel case in name of json attributes
- always generate all json attributes (even empty ones)

PR #177, opened 2017-01-20
Escaping input from perun file

sustr4:
This, hopefully, addresses the risk of bash command injection through Perun output.


sustr4:
Found a bug, will fix ASAP.

sustr4:
Sorry, false alarm. It was only affecting Pull Request #179

PR #188, opened 2017-03-07
Optional location of destination for authorized_keys

balcirakpeter:
Optional location of DST_FILE in process-sshkeys-root added.
Pre script added.


licehammer:
Merged with small modification (for details see 43ef56c7e43aa80c431bedf476b9bfad275a8509)

PR #117, opened 2016-09-21
Change of o365 gen script

stavamichal:

PR #107, opened 2016-09-05
Added O365 service

Laliska:

PR #189, opened 2017-03-07
O365 final

stavamichal:
Prototype of O365 Service


stavamichal:
Final pull request merged from base of #107 and changes of #117.

zlamalp:
We need to explicitly state, that it's prototype service for CzechGlobe, it works with small instance of O365.

PR #206, opened 2017-04-19
Add process of working with duplicit CA+DN identities

stavamichal:
- this behavior is just in generation script for voms
 - same for voms and voms_dirac services
 - for Perun is combination of CA+DN unique (case-sensitive)
 - for Voms is combination of CA+DN without trailing whitespaces
   (everywhere in the combination) and case-insensitive
 - this commit try to map duplicits of such CA+DN from perun to voms and
   working with such situation (only last CA+DN with duplicity is saved
   for one such user record)

### DIFF
--- a/gen/voms
+++ b/gen/voms
@@ -8,7 +8,7 @@ use Text::Unidecode;
 use XML::Simple;
 
 local $::SERVICE_NAME = "voms";
-local $::PROTOCOL_VERSION = "3.1.0";
+local $::PROTOCOL_VERSION = "3.1.1";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
@@ -25,6 +25,7 @@ our $A_G_VOMS_ROLES;    *A_G_VOMS_ROLES =      \'urn:perun:group:attribute-def:d
 our $STATUS_VALID;      *STATUS_VALID =        \'VALID';
 
 my $struc = {};
+my $uniquenessMapping = {};
 
 #resource one by one
 foreach my $resourceData ($data->getChildElements) {
@@ -48,6 +49,7 @@ foreach my $resourceData ($data->getChildElements) {
 		#group members one by one
 		foreach my $memberData (($groupData->getChildElements)[1]->getChildElements) {
 			my %memberAttributes = attributesToHash $memberData->getAttributes;
+			my $memberUniqueIdentifier;
 			#skip member if his status is not valid
 			next unless $memberAttributes{$A_USER_STATUS} eq $STATUS_VALID;
 			#get mail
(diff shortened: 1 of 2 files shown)


gen/voms
```diff
@@ -48,27 +49,47 @@ foreach my $resourceData ($data->getChildElements) {
...
 			next unless $memberAttributes{$A_USER_STATUS} eq $STATUS_VALID;
 			#get mail
 			my $email = $memberAttributes{$A_USER_MAIL};
 
 			#each DN of user is separate instance of user in voms
 			#skip users with no certificates
-			foreach my $subjectDN (keys %{$memberAttributes{$A_USER_CERT_DNS}}) {
+			foreach my $subjectDN (sort keys %{$memberAttributes{$A_USER_CERT_DNS}}) {
+				#set uniqueIdentifier for member (his first certificate DN+CA)
+				unless($memberUniqueIdentifier) { $memberUniqueIdentifier = $subjectDN . "---------------" .  $memberAttributes{$A_USER_CERT_DNS}{$subjectDN}; }
```
sustr4: Wouldn't it be more elegant to get rid of the arbitrary separator and have the `$uniquenessMapping` array two-dimensional? As in `$uniquenessMapping->{$CA}->{$DN}`.

gen/voms
```diff
@@ -48,27 +49,47 @@ foreach my $resourceData ($data->getChildElements) {
...
 				#create new member if not exists in VO yet
 				if(!defined($struc->{$voShortName}->{'users'}->{$uniqueVomsUser})) {
 					$struc->{$voShortName}->{'users'}->{$uniqueVomsUser}->{'email'} = $email;
 					$struc->{$voShortName}->{'users'}->{$uniqueVomsUser}->{'CA'} = $CADN;
 					$struc->{$voShortName}->{'users'}->{$uniqueVomsUser}->{'DN'} = $subjectDNWithoutPrefix;
 					$struc->{$voShortName}->{'users'}->{$uniqueVomsUser}->{'roles'} = {};	
 					$struc->{$voShortName}->{'users'}->{$uniqueVomsUser}->{'groups'} = {};
+				} else {
+					#use the last record of DN and CA we get (it is probably the newest one)
+					$struc->{$voShortName}->{'users'}->{$uniqueVomsUser}->{'CA'} = $CADN;
```
sustr4: It looks like these two assignments are done in both branches of this `if` statement. Shouldn't we just call them **outside** the `if..else` block?

stavamichal: You are right. Fixed.

gen/voms
stavamichal: It is just a unique identifier like "id" but bigger. Separator there is just cosmetic for purpose of better code reading. I can remove it and it still be functional. I don't think we need a two-dimensional array instead, there is probably no advantage in it.

sustr4: OK. it's just cosmetics. Since it's only an ephemeral structure it's of little concern.

PR #220, opened 2017-05-26
Add coffeeroom service

inkuzmin:


zlamalp:
Hi @inkuzmin, thanks for the pull-request. I believe @tauceti2 is in touch with you, we will review this once it's complete (generating script part).

inkuzmin:
Hello, @zlamalp! Thank you for your reply! Yes, this is exactly the case, we are testing it with @tauceti2 now.

zlamalp:
Hi, @tauceti2, @inkuzmin, how this ended up ? Was the script finished / tested / deployed ? Thank you.

inkuzmin:
Hello, @zlamalp! Thank you, we tested and deployed it. I've committed the `process-coffeeroom.sh`, currently this version is used to keep groups' data updated.

slave/process-coffeeroom/changelog
```diff
@@ -0,0 +1,5 @@
+perun-slave-process-coffeeroom (1.0.0) stable; urgency=low
```
zlamalp: Please start package version at 3.1.0 , it's related to perun version. We are in a transition to more proper versioning through perun.

inkuzmin:
Hello @zlamalp! Thank you for correcting me. I have updated a version and added dependencies.
As for `users.csv` and `groups.csv` it's got to be some mistake, I fixed an extension to the `scim`.

zlamalp:
Hi @inkuzmin, thank you for the update. Just one last thing. I still need to know, how are files ``users.scim`` and ``groups.scim`` generated and transfered to destinations. Since I can see on Elixir Perun instance, there is a service called "scim", which generates data and send them to facility coffeeroom. But your slave script is called "coffeeroom". If you are using our standard way of provisioning, name of the service must match in all three parts (gen/send/slave).

I don't want to rename the service to "scim", since it's technically the data exchange format, not specific service. What do you do with the data in a scim format depends on each service. Also we want more services in a future to use scim, if possible.

So what is expected result or solution? Did you consult it with @tauceti2 ? Maybe you should just copy&paste gen/send scripts for the scim and just update service/file name to coffeeroom. If you deploy this version of a script, it won't work without some hack to peruns main slave script.

zlamalp:
Hi @inkuzmin, I talked with @tauceti2 and created pull #265 with current implementation of scim gen/send scripts renamed and fixed for coffeeroom service.

inkuzmin:
Hello @zlamalp! Great, thank you!

PR #222, opened 2017-06-01
Fix error report content in process-voms

sustr4:
On certain occasions `list-groups` and might perhaps pass but following calls fail. Use the most recent error output in reports. 
This is a quick fix. May merge.


stavamichal:
Please also modify changelog of this service to new version. Thank you.

PR #224, opened 2017-06-07
Implemented cross-distro compatible locking in slave scripts

zlamalp:
- Base package will install also /usr/lib/tmpfiles.d/perun.conf
  which defines new folder /var/lock/perun/, which is generally
  writable for all users.
  This solve case, when slave script was running under normal user
  and didn't have privileges to create lock in /var/lock/ on CentOS host.
- Locking directory changed to /var/lock/perun/


zlamalp:
For some reason, /var/lock/perun/ is not created on deb7. Will investigate soon.

zlamalp:
Since this would require systemd to be installed on destination machine, more compatible solution is in #228. Basically we allow custom lockdir config and if not writable by current user, script fallbacks to /tmp.

PR #228, opened 2017-06-15
Check if lock dir is writable in perun-base package

zlamalp:
- In generic slave script perun from perun-base package, check if directory
  for locks is writable by current user. If not, use /tmp as base for lock dir path.
  This should be most compatible way to handle locks for all users and OS.

  Default lock dir is /var/lock/, but can be set custom in /etc/perunv3.conf.

### DIFF
--- a/slave/base/bin/perun
+++ b/slave/base/bin/perun
@@ -114,6 +114,13 @@ function catch_error {
 }
 
 function create_lock {
+
+	# test if lock dir is writable !
+	if ! [ -w $LOCK_DIR ] ; then
+		# If not - redirect lock to /tmp as most compatible non-persistent place for locks.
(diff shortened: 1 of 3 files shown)


slave/base/bin/perun
```diff
@@ -15,6 +15,12 @@ SERVICE_WHITELIST=()
 DNS_ALIAS_WHITELIST=( `hostname -f` )
 FACILITY_WHITELIST=()        # from which facilities this host accept configuration
 
+# check if lock dir is writable
+if ! [ -w $LOCK_DIR ] ; then
+    # If not - redirect lock to /tmp as most compatible non-persistent place for locks.
+    LOCK_DIR="/tmp";
```
licehammer: Use \t for indentation, please.

PR #229, opened 2017-06-16
New SCIM service

tauceti2:
SCIM sends information about users and group and memberships.

### DIFF
--- a/gen/scim
+++ b/gen/scim
@@ -1,0 +1,175 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use perunServicesInit;
+use perunServicesUtils;
+use Perun::Agent;
+use Perun::GroupsAgent;
+use open qw/:std :utf8/;
+use JSON::XS;
+use utf8;
+
+local $::SERVICE_NAME = "scim";
+local $::PROTOCOL_VERSION = "1.0.0";
+my $SCRIPT_VERSION = "1.0.0";
+
+perunServicesInit::init;
+my $DIRECTORY = perunServicesInit::getDirectory;
+my $data = perunServicesInit::getDataWithGroups;
+
+#forward declaration
+sub processUsers;
+sub processGroups;
+sub processMemberships;
+
+#Constants
+our $A_USER_ID;                   *A_USER_ID =               \'urn:perun:user:attribute-def:core:id';
+our $A_USER_STATUS;               *A_USER_STATUS =           \'urn:perun:member:attribute-def:core:status';
+our $A_USER_EMAIL;                *A_USER_EMAIL =            \'urn:perun:user:attribute-def:def:preferredMail';
+our $A_USER_LOGIN;                *A_USER_LOGIN =            \'urn:perun:user_facility:attribute-def:virt:login';
+our $A_USER_D_NAME;               *A_USER_D_NAME =           \'urn:perun:user:attribute-def:core:displayName';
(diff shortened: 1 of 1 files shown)


gen/scim
```diff
@@ -0,0 +1,176 @@
...
+
+our $STATUS_VALID;                *STATUS_VALID =            \'VALID';
+our $STATUS_EXPIRED;              *STATUS_EXPIRED =          \'EXPIRED';
+our $STATUS_SUSPENDED;            *STATUS_SUSPENDED =        \'SUSPENDED';
+
+our $userStruc = {};
+our $u_status = {};
+our $u_email = {};
+our $u_name = {};
+our $u_login = {};
```
licehammer: Please use camelCase.
There is no need to initialize these variables as empty hash reference.
Do you really need these variable to be global?

gen/scim
```diff
@@ -0,0 +1,176 @@
...
+use warnings;
+use perunServicesInit;
+use perunServicesUtils;
+use Perun::Agent;
+use Perun::GroupsAgent;
+use open qw/:std :utf8/;
+use JSON::XS;
+use utf8;
+
+local $::SERVICE_NAME = "scim";
```
licehammer: Are you sure about naming this service "scim"?  How would slave script for this service look like?

gen/scim
```diff
@@ -0,0 +1,176 @@
...
+        foreach my $groupData (($resourceData->getChildElements)[0]->getChildElements) {
+                my $groupMembersLogins = processGroups $groupData;
+        }
+}
+
+# PREPARE USERSDATA TO JSON
+my @users;
+foreach my $uid (sort keys %$userStruc) {
+        my $user = {};
+        $user->{"id"} = $uid;
```
licehammer: Do we want to reveal internal Perun ID to other systems?

gen/scim
```diff
@@ -0,0 +1,176 @@
...
+                        $userStruc->{$uid}->{$u_status} = $status;
+                } elsif ($memberStatus eq $STATUS_SUSPENDED && $status eq $STATUS_VALID){
+                        # change from SUSPENDED to VALID
+                        $userStruc->{$uid}->{$u_status} = $status;
+                } elsif ($memberStatus eq $STATUS_SUSPENDED && $status eq $STATUS_EXPIRED){
+                        # change from SUSPENDED to EXPIRED
+                        $userStruc->{$uid}->{$u_status} = $status;
+                }
+        }
+        else{
```
licehammer: Missing space before '{'

gen/scim
```diff
@@ -0,0 +1,176 @@
...
+                }
+        }
+}
+
+## creates structure for memberships
+sub processMemberships {
+        my ($gid, $uid) = @_;
+
+        unless(exists $membershipStruc->{$gid}->{$uid}) {
+                $membershipStruc->{$gid}->{$uid} = {};
```
licehammer: Is membershipStruc used for anything?

gen/scim
tauceti2: The slave already exists. It just drop files at the target system and external script is usually getting those files for processing.

gen/scim
tauceti2: actually this is the only way, how you can link memberships and users

gen/scim
tauceti2: yes for printing users membership into the groups.scim

licehammer: My bad, I overlooked that part of code.

PR #246, opened 2017-08-01
New slave for SCIM service

tauceti2:


slave/process-scim/bin/process-scim.sh
```diff
@@ -0,0 +1,56 @@
...
+	diff_mv "${FROM_PERUN_USERS}" "${DST_FILE_USERS}"
+	user_diff=$?
+	diff_mv "${FROM_PERUN_GROUPS}" "${DST_FILE_GROUPS}"
+	group_diff=$?
+	
+	if [ $user_diff -eq 0 ] || [ $group_diff -eq 0 ]; then
+		log_msg I_CHANGED
+		# if script for processing SCIM data exists and is runnable, then run it
+		if [ -x ${PROCESS_SCRIPT} ]; then
+			$DST_KYPO_IMPORT
```
stavamichal: This is probably not correct, there is no definition of this variable above.

slave/process-scim/bin/process-scim.sh
```diff
@@ -0,0 +1,56 @@
...
+			exit $?
```
stavamichal: If you use exit this way, you can have inconsistency with same error status but different meaning of it, because you already have errors with return codes 100+ there and exec scipt can return also these codes but with different meanings. Just think about it.

slave/process-scim/changelog
```diff
@@ -0,0 +1,5 @@
+perun-slave-process-scim (3.0.0) stable; urgency=low
+
+  * Added service perun-slave-scim
```
stavamichal: I know there is already "scim" in the name of service, but i would prefer to have purpose of this slave script in this message too, please.

slave/process-scim/lib/process-scim.example
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python3
+'''
+EXAMPLE of python script for processing SCIM files from Perun
+'''
+
+USERS_FILE = '/var/spool/perun/scim-service-users.scim'
+GROUPS_FILE = '/var/spool/perun/scim-service-grops.scim'
```
stavamichal: These two files should be taken from input of this script (in execution of this script on bash side), instead of duplicate their path in bash and in python.

slave/process-scim/bin/process-scim.sh
```diff
@@ -0,0 +1,56 @@
...
+#
+#author:  Michal Prochazka
+#date:	  2017-08-01
+#
+#
+
+PROTOCOL_VERSION='1.0.0'
+
+function process {
+	DST_DIR="/var/spool/perun/"
```
stavamichal: I don't think you need this directory. You can send them to the executed script as parameters.

slave/process-scim/lib/process-scim.example
```diff
@@ -0,0 +1,59 @@
...
+		#This removes prefix of parent group exported from perun, that is not suitable
+		tmpGroup.name = tmpGroup.name.split(':')[-1]
+
+	groups_list.append(tmpGroup)
+	
+	for i in item['members']:
+		tmpUserInGroup = UserInGroup()
+		tmpUserInGroup.user_external_id = int(i['userId'])
+		tmpUserInGroup.group_external_id = int(item['id'])
+		usersInGroups_list.append(tmpUserInGroup)
```
stavamichal: I can't find usersInGroups_list variable anywhere in the script above.

slave/process-scim/bin/process-scim.sh
```diff
@@ -0,0 +1,56 @@
...
+	diff_mv "${FROM_PERUN_GROUPS}" "${DST_FILE_GROUPS}"
+	group_diff=$?
+	
+	if [ $user_diff -eq 0 ] || [ $group_diff -eq 0 ]; then
+		log_msg I_CHANGED
+		# if script for processing SCIM data exists and is runnable, then run it
+		if [ -x ${PROCESS_SCRIPT} ]; then
+			$DST_KYPO_IMPORT
+			exit $?               
+		fi
```
licehammer: Please add else branch which raises error due to $PROCESS_SCRIPT is not executable.

slave/process-scim/bin/process-scim.sh
tauceti2: I want to use diff_mv, so I need to have a local copy.

slave/process-scim/bin/process-scim.sh
```diff
@@ -0,0 +1,56 @@
...
+	I_NOT_CHANGED=(0 "${DST_FILE_USERS} and ${DST_FILE_GROUPS} has not changed")
+	E_DIR_NOT_EXISTS=(100 "${DST_DIR} does not exists")
+	E_DIR_NOT_WRITABLE=(101 "${DST_DIR} is not writable for user ${CURRENT_USER}")
+	
+	FROM_PERUN_USERS="${WORK_DIR}/users.scim"
+	FROM_PERUN_GROUPS="${WORK_DIR}/groups.scim"
+
+	if [ ! -d "${DST_DIR}" ]; then
+		log_msg	E_DIR_NOT_EXISTS
+		exit 1
```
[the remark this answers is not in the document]
tauceti2: But the result of test command is true, so then the log_msg will return 0.

slave/process-scim/bin/process-scim.sh
tauceti2: Right, that variable is set in slave configuration file, I will add documentation here.

slave/process-scim/bin/process-scim.sh
stavamichal: In this case this is ok.
But using diff_mv this way you have different problem: "What if script makes diff_mv with not-null return code and then ends with error? Another run of script will not change anything, because diff_mv will be without any change."

slave/process-scim/lib/process-scim.example
```diff
@@ -0,0 +1,59 @@
...
+json_users.close()
+
+json_groups = open(GROUPS_FILE)
+groups_data = json.load(json_groups)
+json_groups.close()
+
+''' PARSING DATA FROM users.scim '''
+for item in users_data:
+	userJSON_ids.append(int(item['id']))	
+	tmpUser = User()
```
licehammer: Pyhton code style recommends using underscores instead of cammelCase. I don't mind either, but seeing combination of both in one file really draws ones attention :-)

stavamichal:
Can you please process all mentioned comments? Thank you.

slave/process-scim/bin/process-scim.sh
tauceti2: Because PROCESS_SCRIPT is our script, then it requires to follow the same rules for return codes.

slave/process-scim/bin/process-scim.sh
tauceti2: If there is an error inside diff_mv, then it raises log_msg with error message, so it ends with exit non 0 value.

slave/process-scim/lib/process-scim.example
tauceti2: This is an example where the data coming from Perun ends. I'll add comment there.

slave/process-scim/bin/process-scim.sh
tauceti2: Fixed. Now it returns 103 if external script fails

slave/process-scim/lib/process-scim.example
tauceti2: Variables are passed to the external script via command line arguments.

slave/process-scim/lib/process-scim.example
tauceti2: Comments added.

slave/process-scim/lib/process-scim.example
tauceti2: Fixed.

slave/process-scim/changelog
tauceti2: Description added.

stavamichal:
@licehammer Can you please check python part of script and check if you are ok with changes? Thank you.

slave/process-scim/bin/process-scim.sh
```diff
@@ -0,0 +1,60 @@
...
+	diff_mv "${FROM_PERUN_USERS}" "${DST_FILE_USERS}"
+	user_diff=$?
+	diff_mv "${FROM_PERUN_GROUPS}" "${DST_FILE_GROUPS}"
+	group_diff=$?
+	
+	if [ $user_diff -eq 0 ] || [ $group_diff -eq 0 ]; then
+		log_msg I_CHANGED
+		# if script for processing SCIM data exists and is runnable, then run it
+		if [ -x ${PROCESS_SCRIPT} ]; then
+			eval ${PROCESS_SCRIPT} ${DST_FILE_USERS} ${DST_FILE_GROUPS}
```
licehammer:
```suggestion
			eval "${PROCESS_SCRIPT}" "${DST_FILE_USERS}" "${DST_FILE_GROUPS}"
```

slave/process-scim/bin/process-scim.sh
```diff
@@ -0,0 +1,60 @@
...
+
+	diff_mv "${FROM_PERUN_USERS}" "${DST_FILE_USERS}"
+	user_diff=$?
+	diff_mv "${FROM_PERUN_GROUPS}" "${DST_FILE_GROUPS}"
+	group_diff=$?
+	
+	if [ $user_diff -eq 0 ] || [ $group_diff -eq 0 ]; then
+		log_msg I_CHANGED
+		# if script for processing SCIM data exists and is runnable, then run it
+		if [ -x ${PROCESS_SCRIPT} ]; then
```
licehammer:
```suggestion
		if [ -x "${PROCESS_SCRIPT}" ]; then
```

slave/process-scim/bin/process-scim.sh
```diff
@@ -0,0 +1,60 @@
...
+	if [ ! -w "${DST_DIR}" ]; then
+		log_msg	E_DIR_NOT_WRITABLE
+	fi
+
+	create_lock
+
+	diff_mv "${FROM_PERUN_USERS}" "${DST_FILE_USERS}"
+	user_diff=$?
+	diff_mv "${FROM_PERUN_GROUPS}" "${DST_FILE_GROUPS}"
+	group_diff=$?
```
licehammer: Common practice in bash is to have variable names in uppercase.

slave/process-scim/rpm.dependencies
```diff
@@ -0,0 +1 @@
+perun-slave-base
```
licehammer: I would expect python3 here.

tauceti2: The slave doesn't have dependency on python. Script written in python is just an example how to use the slave.

slave/process-scim/bin/process-scim.sh
tauceti2: Thanks fixed.

slave/process-scim/bin/process-scim.sh
tauceti2: Thanks fixed.

PR #251, opened 2017-08-25
Update of kypo portal slave to remove tmp files if ends with error

frhrdina:
Update of kypo portal slave to remove tmp files if ends with error

### DIFF
--- a/slave/process-kypo-portal/bin/process-kypo_portal.sh
+++ b/slave/process-kypo-portal/bin/process-kypo_portal.sh
@@ -39,8 +39,9 @@ function process {
 		python3 $DST_KYPO_IMPORT
 	fi
 
+	# If python import script ends with error, tmp files are deleted
 	RETVAL=$?
-	if [ $RETVAL -eq 1 ]; then
+	if [ $RETVAL -ne 0 ]; then
 		rm $DST_FILE_USERS $DST_FILE_GROUPS
 	fi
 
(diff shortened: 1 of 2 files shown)


slave/process-kypo-portal/bin/process-kypo_portal.sh
```diff
@@ -31,11 +31,16 @@ function process {
 	diff_mv "${FROM_PERUN_GROUPS}" "${DST_FILE_GROUPS}"
 	group_diff=$?
 	
-	if [ $user_diff -eq 0 ] || [ $group_diff -eq 0 ]; then
+	if [ $user_diff -ne 0 ] && [ $group_diff -ne 0 ]; then
+		log_msg I_NOT_CHANGED
```
kouril: you should do exit 0 here. Otherwise the flow is flow is a bit more complex (with the 'else') and the files in /tmp get always deleted.

slave/process-kypo-portal/bin/process-kypo_portal.sh
```diff
@@ -31,11 +31,16 @@ function process {
...
-		exit $?               
-	else
-		log_msg I_NOT_CHANGED
 	fi
+
+	if [ $? -eq 1 ]; then
+		rm $DST_FILE_USERS $DST_FILE_GROUPS
+	fi
+
+	exit $?
```
kouril: the return code of rm might be returned here!

PR #252, opened 2017-08-31
Implemented ad_mu service for filling O365 AD

zlamalp:
- Add and update all users with ZAM/STU relation on MU.
- Add / update groups, manage their members directly.
- Special grace period behavior for license groups.
  (employees and students are kept for 30 days in previous
  upper group, also people, which were once students are
  put to alumni, instead of complete removal).

- Updated ADConnector.pm lib to allow reading of OUs from
  Perun and AD.

### DIFF
--- a/gen/ad_mu
+++ b/gen/ad_mu
@@ -436,6 +436,8 @@ for my $login (@logins) {
 	}
 	print FILE "MailNickName: " . $login . "\n";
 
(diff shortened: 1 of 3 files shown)


stavamichal:
This functionality has been changed so we will create a new pull request soon. This can be closed.

PR #258, opened 2017-10-03
Add new destination type to generic send

stavamichal:
- add email type just for notifying user that his way is not
   implemented yet
 - add url type with working code using curl to update file to web
   url

### DIFF
--- a/send/generic_send
+++ b/send/generic_send
@@ -12,8 +12,27 @@ FACILITY_NAME=$1
 DESTINATION=$2
 DESTINATION_TYPE=$3
 KEY_PATH="`echo ~`/.ssh/id_rsa"
-TRANSPORT_COMMAND="ssh -o StrictHostKeyChecking=no -o GSSAPIAuthentication=no -o GSSAPIKeyExchange=no -o ConnectTimeout=5  -i $KEY_PATH"
 
+#predefined different types of destination
+DESTINATION_TYPE_URL="url"
+DESTINATION_TYPE_EMAIL="email"
+DESTINATION_TYPE_HOST="host"
+DESTINATION_TYPE_USER_HOST="user@host"
+DESTINATION_TYPE_USER_HOST_PORT="user@host:port"
+
+#if there is no destination type, use default 'host'
+if [ -z "$DESTINATION_TYPE" ]; then
+	DESTINATION_TYPE=$DESTINATION_TYPE_HOST
+fi
+
+#choose transport command, only url type has different transport command at this moment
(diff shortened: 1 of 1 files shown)


send/generic_send
```diff
@@ -83,12 +92,25 @@ if [ $? -ne 0 ]; then
 	exit 255
 fi
 
+TEMPORARY_FILE="/tmp/${SERVICE_NAME}-$$.tar"
```
tauceti2: It will be good to delete TEMPORARY_FILE when the script ends.

send/generic_send
```diff
@@ -83,12 +92,25 @@ if [ $? -ne 0 ]; then
...
+	tar -c -f "$TEMPORARY_FILE" -C "$SERVICE_FILES_DIR"  --exclude="_destination" .  -C "$TMP_HOSTNAME_DIR" . 
 fi
+
+if [ $? -ne 0 ]; then
+	echo "Can't archive service files to $TEMPORARY_FILE" >&2
+	exit 255
+fi
+
+if [ "$DESTINATION_TYPE"="url" ]; then
+	timeout -k $TIMEOUT_KILL $TIMEOUT $TRANSPORT_COMMAND --data-binary @$TEMPORARY_FILE $HOST
```
licehammer: curl can read data from STDIN using following syntax. Therefore there is no need to create extra file (\$TEMPORARY_FILE)
```bash
curl -i -H Content-Type:application/tar -f -X PUT --data-binary -
```

stavamichal:
@licehammer I fixed that by your recommendation, can you please check it?

send/generic_send
```diff
@@ -59,6 +59,16 @@ case $DESTINATION_TYPE in
 		HOSTNAME=`echo $HOST | sed -e 's/^.*@//'`
 		PORT=`echo $DESTINATION | sed -e 's/^.*://'`
 		;;
+	url)
+		HOSTNAME="$DESTINATION"
+		HOST="$DESTINATION"
+		TRANSPORT_COMMAND="curl -i -H Content-Type:application/tar -f -X PUT --data-binary - "
```
zlamalp: If we override TRANSPORT_COMMAND for service on line 18 (by including conf/sh file), it gets overwritten again by this in case of URL destination type. This definition for URL destination type should be above line 18. Somebody might want to pass certificates or login/pass for this connection and I think it's part of curl command.

stavamichal:
I reworked it to support overriding for transport command. Please check it now.

PR #277, opened 2017-11-14
Add missing versions of perun-base dependency for k5login and fs-home

stavamichal:
- there is a new method needed for k5login and for fs-home so
   dependency need to have set proper version


licehammer:
Merged locally. I've added a comment to commit message about this being addition to dea9364fd1740735feda3cce104da310772542f6

PR #279, opened 2017-11-16
Add authentication by certificate to URL type of destination

stavamichal:
- if there are key and cert files on predefined paths
   "/etc/perun/ssl/perun-send.key|pem" then use them for URL type of
   destination

### DIFF
--- a/send/generic_send
+++ b/send/generic_send
@@ -13,6 +13,9 @@ DESTINATION=$2
 DESTINATION_TYPE=$3
 KEY_PATH="`echo ~`/.ssh/id_rsa"
 
+PERUN_CERT="/etc/perun/ssl/perun-send.pem"
+PERUN_KEY="/etc/perun/ssl/perun-send.key"
+
 #predefined different types of destination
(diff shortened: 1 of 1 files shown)


send/generic_send
```diff
@@ -13,6 +13,9 @@ DESTINATION=$2
 DESTINATION_TYPE=$3
 KEY_PATH="`echo ~`/.ssh/id_rsa"
 
+PERUN_CERT="/etc/perun/ssl/perun-send.pem"
+PERUN_KEY="/etc/perun/ssl/perun-send.key"
```
zlamalp: Are you sure we won't need different certificates per service or even destination ? ``/etc/perun/ssl/`` is usually reserved for server cert (apache/tomcat), what about ``/etc/perun/services/$serviceName/perun-send.pem or .key``.
Everything can be overriden by content of ``/etc/perun/services/${SERVICE_NAME}/${SERVICE_NAME}.conf`` on service level so it doesn't really matter. Just sharing opinions.

PR #307, opened 2018-02-02
Modify service o365_mu

stavamichal:
- change in gen script, for forwarding is now used default address
 if forwarding email is empty. If disable for forwarding is set, then
 we will use empty value for forwarding.


gen/o365_mu
```diff
@@ -135,7 +139,12 @@ sub processMember {
 	my %memberAttributes = attributesToHash $memberData->getAttributes;
 
 	my $UCO = $memberAttributes{$A_UF_LOGIN};
+
+	my $disableForward = $memberAttributes{$A_UF_DISABLE_MAIL_FORWARD};
 	my $mailForward = $memberAttributes{$A_UF_O365_MAIL_FORWARD};
```
licehammer: If $disableForward is true $mailForward should be empty.

stavamichal: Fixed.

stavamichal:
I removed "skipping empty lines" from commit, because I can't prove it is necessary for correct behavior of script.

PR #309, opened 2018-02-14
New gen script for service firewall

stavamichal:
- this service takes all rules from all resources on the facility
   and for every such rule take all IPAddresses of assigned users.
   These infromations then print to the json file with the same name as
   service itself.


stavamichal:
All mentioned problems were fixed.

PR #315, opened 2018-02-27
New slave script for service o365_contacts_export

stavamichal:
- this new service process file with exported emails, copies it to
   the expected destination (can be changed by pre-scripts) and then
   executes expected command (also can be changed by pre-scripts) for
   this new file. If command fails, this slave script also fails but
   with it's own error code (does not preserve error code from the target
   script).

### DIFF
--- a/slave/process-o365-contacts-export/changelog
+++ b/slave/process-o365-contacts-export/changelog
@@ -1,0 +1,5 @@
+perun-slave-process-o365-contacts-export (3.0.0) stable; urgency=low
+
+  * Package for exporting Perun managed o365 emails
+
+ -- Michal Stava <stavamichal@gmail.com>  Tue, 27 Feb 2018 10:07:00 +0100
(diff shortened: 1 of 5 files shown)


slave/process-o365-contacts-export/changelog
```diff
@@ -0,0 +1,5 @@
+perun-slave-process-o365-contacts-export (3.0.0) stable; urgency=low
+
+  * Package for exporting o365 conctacts from Perun to o365
```
licehammer: Please replace with following:
"Package for exporting Perun managed o365 emails"

slave/process-o365-contacts-export/short_desc
```diff
@@ -0,0 +1 @@
+Package for perun service - export of emails for o365 purpose
```
licehammer: Please replace with following:
"Package for perun service - export Perun managed o365 emails"

stavamichal:
Both texts were replaced.

PR #316, opened 2018-03-14
Extend service googleGroups to use team drives

stavamichal:
- add team drives to gen service (use drive name from resource, more
   resources with same name are merged together), new file with all team
   drives - google_groups_team_drives.csv
 - add team drives to send script - call java application with option
   'teamDrives' and input file with team drives

### DIFF
--- a/gen/google_groups
+++ b/gen/google_groups
@@ -14,17 +14,23 @@ my $DIRECTORY = perunServicesInit::getDirectory;
 my $data      = perunServicesInit::getDataWithGroups;
 
 #Constants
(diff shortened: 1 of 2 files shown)


zlamalp:
Just generic comment. TeamDrives require users mail addresses, while perun attribute for google login provides just number identification (identity), so we must get mail address from proper UserExtSource attribute first, then include it to this service. I'm not sure we would be allowed to "read user" from Google API in java connector even if we know his ID (user is not in our domain, but generic google). Actually there was Drive API function in v2, but current API v3 doesn't have it.

PR #318, opened 2018-03-22
Exit with correct return code for service o365_mu

stavamichal:
- we need to exit with correct return code when there is an error in
   o365_mu service
 - use global return code for whole script. If any thread ends with
   error, whole script will ends with error in the end.
 - also if parsing of file return empty line (without correct
   identifier), script will also ends with not-null return code

### DIFF
--- a/send/o365_mu_process.pl
+++ b/send/o365_mu_process.pl
@@ -33,7 +33,6 @@ my $FILE_GROUPS_LOCK :shared;
 
 #number of worker threads
 my $THREAD_COUNT = 10;
-our $GLOBAL_RETURN_CODE=0;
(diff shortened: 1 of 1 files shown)


send/o365_mu_process.pl
```diff
@@ -33,6 +33,7 @@
 
 #number of worker threads
 my $THREAD_COUNT = 10;
+our $GLOBAL_RETURN_CODE=0;
```
licehammer: Let's keep this local (my). There is no need to access this variable from other imported files/libraries.

stavamichal:
I hope it is fixed now.

PR #326, opened 2018-04-12
Use new group_resource attributes in voms services

stavamichal:
- instead of "group" specific attributes "vomsGroupName" and
   "vomsRoles" we want to use same attribute for namespace
   "group_resource". This help to use one group for more than one
   resource with different settings.


stavamichal:
Need to wait for correct modules from https://github.com/CESNET/perun/pull/1724 !

PR #329, opened 2018-05-28
New service for access to Foreman

stavamichal:
- gen and send script for access to Foreman


stavamichal:
We won't need this service. I am closing this PR.

PR #330, opened 2018-06-04
Process timeout in scripts

stavamichal:
- return code 124 means that limited time has run out and script was
   killed by the command "timeout"
 - in the situations when time has run out, we want to know it on the error
   output

### DIFF
--- a/send/topdesk
+++ b/send/topdesk
@@ -28,6 +28,11 @@ timeout -k $TIMEOUT_KILL $TIMEOUT $TRANSPORT_COMMAND
 
 ERR_CODE=$?
 
-echo "Communication with slave script ends with return code: $ERR_CODE" >&2
+#Error code 124 means - timeouted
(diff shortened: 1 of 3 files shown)


send/topdesk
```diff
@@ -28,6 +28,11 @@ timeout -k $TIMEOUT_KILL $TIMEOUT $TRANSPORT_COMMAND
 
 ERR_CODE=$?
 
-echo "Communication with slave script ends with return code: $ERR_CODE" >&2
+#Error code 124 means - timeouted
+if [ $ERR_CODE -e 124 ]; then
+	echo "Communication with slave script was timeouted with return code: $ERR_CODE (Warning: this error can mask original error 124 from slave script!)" >&2
+else
+  echo "Communication with slave script ends with return code: $ERR_CODE" >&2
```
licehammer: Indent.

stavamichal: Fixed.

PR #339, opened 2018-06-27
Service ad_mu - add limit of memberships in group

licehammer:
Group memberships from AD are transfered to Microsoft o365 cloud using AD
Connect which have a limitation of 49999 members in a group.

Only group we can realistically hit this limitation is Students license
group containing all users who are allowed to get students license. To
overcome this limitation, Student2 group is created and all users over
49999 from Student group are stored into Students2 group instead.


licehammer:
This change was already tested in production environment.

PR #354, opened 2018-08-22
Service ad_mu - fix emptying of a group

licehammer:
There were switch parameters in LDAP update call, which leads to
following error:
"Can't locate object method "update" via package "Net::LDAPS" at ./ad_mu
line 765."

This part of code was used only to empty groups that are no longer
managed by Perun.


licehammer:
The code was tested on a production instance.

PR #355, opened 2018-08-22
ad-mu - disable removing stale lock and lock configuration

licehammer:
Successfully tested on a production instance.

### DIFF
--- a/send/ad_mu
+++ b/send/ad_mu
@@ -32,7 +32,7 @@ use Net::LDAP::Control::Paged;
 use Net::LDAP::Constant qw( LDAP_CONTROL_PAGED );
 use File::Copy;
 use Date::Calc qw/ Today Delta_Days Add_Delta_Days Date_to_Days /;
-use LockFile::Simple qw(lock trylock unlock);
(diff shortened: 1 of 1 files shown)


send/ad_mu
```diff
@@ -368,8 +368,9 @@ my $lock = ScriptLock->new($facility_name . "_" . $service_name . "_" . $namespa
 ($lock->lock() == 1) or die "Unable to get lock, service propagation is already running or previous run failed.";
 
 # check for errors in last run - if there were an error this lock cannot be obtained
+my $lockManager = LockFile::Simple->make( -hold => 0, -delay => 30, -max => 2);
```
zlamalp: So, the difference between our ``ScriptLock.pm`` and above is, that here you can specify delay and how many times we try to get lock before exiting the script, right ?

licehammer: Right, but the main difference is that ScriptLock will unlock itself on the exit of a script. LockFile::Simple never unlock itself if you have the configuration right, which is what we need to check for an error from the previous run.

PR #362, opened 2018-10-04
Fixed k5login_root behaviour without data

balcirakpeter:
- Problem: When there were no groups assigned on resource
           or assigned groups had no valid members, gen script didnt't generate any file.
           Slave script couldn't proccess that.
- Change:  Gen script edited, so it's generating empty files in these situations
- Result:  Slave script is able to proccess empty files so it works without errors.

### DIFF
--- a/gen/k5login_root
+++ b/gen/k5login_root
@@ -20,7 +20,7 @@ our $A_USER_STATUS;          *A_USER_STATUS =         \'urn:perun:member:attribu
 our $A_GROUP_DESTINATIONS;   *A_GROUP_DESTINATIONS =  \'urn:perun:group:attribute-def:def:listOfDestinations';
 
 my $sortingFunction = getAttributeSorting($A_PRINCIPAL, 1);
-my %outputByDestination = ();
+my %outputByDestination = ('all' => undef);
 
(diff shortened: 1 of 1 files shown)


gen/k5login_root
```diff
@@ -45,7 +45,7 @@ foreach my $resourceData (@resourcesData) {
 	local $, = "\n";
 	local $\ = "\n";
 
-	for my $destination (keys %outputByDestination) {
+	for my $destination (@generatedDataDestination) {
```
stavamichal: This behavior will use only destinations from the last processed group, because it will always replace content of the array in the for-each cycle.

balcirakpeter: Fixed

gen/k5login_root
```diff
@@ -39,6 +43,11 @@ foreach my $resourceData (@resourcesData) {
 	}
 }
 
+foreach my $destination (@allDestinations) {
+	if (!exists $outputByDestination{$destination}) { $outputByDestination{$destination} = undef }
+}
```
stavamichal: I would move this part of code into foreach through groups. Because is hard to understand from code why is it right here even if it works ok.

PR #368, opened 2018-10-17
perunDataGenerator module

balcirakpeter:
- perl module with subroutine generating all user and user_facility attributes (for all users on resource)
  which are required for given service.

### DIFF
--- a/gen/perunServicesInit.pm
+++ b/gen/perunServicesInit.pm
@@ -71,7 +71,12 @@ sub init {
 	} else {
 		unless(defined $facilityId) { die "ERROR: facilityId required"; }
 
-		$agent = Perun::Agent->new('jsonsimple');
+		my $jsonFormat = $::JSON_FORMAT || "jsonsimple";
+		unless($jsonFormat =~ /^json(simple)?$/) {
+			die 'Unsupported json format. Set $::JSON_FORMAT to any of supported formats. Only "json" and "jsonsimple" are supported.';
+		}
+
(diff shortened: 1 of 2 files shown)


gen/perunServicesInit.pm
```diff
@@ -192,4 +195,12 @@ sub createFacilityNameFile($) {
 	close FACILITY_NAME_FILE;
 }
 
+sub setStandardJsonType {
+	$JSON_TYPE='json';
```
licehammer: I'm not happy with a function that changes a global variable. That is a bad practice.

Historically, the perunServicesInit module uses global variables to read data from a gen script. That's also not nice, but it's somehow acceptable for such a small, one-purpose program.

Therefore, please use a global variable as well or rewrite this whole module using the object oriented approach. So usage would be like following.
```perl
my $psi = perunServicesInit->new();
$psi->setJsonType(Perun::Agent::$JSON_SIMPLE)
```

stavamichal: Ok, so we will use global variable, because changing the whole process to be object oriented will be pain at this moment.

gen/perunServicesInit.pm
```diff
@@ -29,6 +29,7 @@ our $DATA_TYPE = {
 	$DATA_TYPE_WITH_VOS => sub {getDataWithVos(@_)},
 };
 
+
```
stavamichal: Is this newline important here?

balcirakpeter: Fixed

PR #372, opened 2018-10-24
fs_scratch_local conditions modification

balcirakpeter:
- modify behavior in fs_scratch_local service to be
  able to continue from it's own failure and also change ACLs on
  scratch directories when there are any differencies

### DIFF
--- a/slave/process-fs-scratch-local/bin/process-fs_scratch_local.sh
+++ b/slave/process-fs-scratch-local/bin/process-fs_scratch_local.sh
@@ -30,21 +30,22 @@ function process {
 
 	# lines contains login\tUID\tGID\t...
 	while IFS=`echo -e "\t"` read LOGIN U_UID U_GID SOFT_QUOTA_DATA HARD_QUOTA_DATA SOFT_QUOTA_FILES HARD_QUOTA_FILES; do
+		SCRATCH_DIR="${SCRATCH_MOUNTPOINT}/${LOGIN}"
 
-		if [ ! -d "${SCRATCH_MOUNTPOINT}/${LOGIN}" ]; then
-			catch_error E_CANNOT_CREATE_DIR  mkdir "${SCRATCH_MOUNTPOINT}/${LOGIN}"
-
-			catch_error E_CANNOT_SET_OWNERSHIP chown "${U_UID}":"${U_GID}" "${SCRATCH_MOUNTPOINT}/${LOGIN}"
-			catch_error E_CANNOT_SET_PERMISSIONS chmod "${UMASK}" "${SCRATCH_MOUNTPOINT}/${LOGIN}"
-
-			# Set quota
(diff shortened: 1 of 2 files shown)


slave/process-fs-scratch-local/bin/process-fs_scratch_local.sh
```diff
@@ -30,21 +30,22 @@ function process {
...
+			log_msg I_DIR_CREATED
+		fi
 
-			# Set quota
-			# setquota name block-softlimit block-hardlimit inode-softlimit inode-hardlimit filesystem
-			#      if [ -x /usr/sbin/setquota ]; then
-			#        # Get $QUOTA_FS
-			#        /usr/sbin/setquota $U_UID $SOFT_QUOTA_DATA $HARD_QUOTA_DATA $SOFT_QUOTA_FILES $HARD_QUOTA_FILES $QUOTA_FS
-			#      fi
+		if [ `stat -L -c "%u:%g" ${SCRATCH_DIR}` != "${U_UID}:${U_GID}" ]; then
```
licehammer:
```suggestion
		if [ "`stat -L -c '%u:%g' ${SCRATCH_DIR}`" != "${U_UID}:${U_GID}" ]; then
```
It's safer to write it this way in case of stat command fails and don't write anything to stdout.

slave/process-fs-scratch-local/bin/process-fs_scratch_local.sh
```diff
@@ -30,21 +30,22 @@ function process {
...
+			catch_error E_CANNOT_SET_OWNERSHIP chown ${U_UID}:${U_GID} "${SCRATCH_DIR}"
+			log_debug "Ownership of ${SCRATCH_DIR} was set to ${U_UID}:${U_GID}"
+		fi
 
-			log_msg I_DIR_CREATED
+		if [ `stat -L -c "%a" ${SCRATCH_DIR}` != "${UMASK}" ]; then
```
licehammer:
```suggestion
		if [ "`stat -L -c '%a' ${SCRATCH_DIR}`" != "${UMASK}" ]; then
```

slave/process-fs-scratch-local/changelog
```diff
@@ -1,3 +1,10 @@
+perun-slave-process-fs-scratch-local (3.1.5) stable; urgency=low
+
+  * Modify service behaviour to be able to continue from it's own failure
```
licehammer: This is not comprehensible. You probably want to say it can recover from a failure in the previous run.

slave/process-fs-scratch-local/changelog
```diff
@@ -1,3 +1,10 @@
...
+  * Change access control list on scratch directory when there are any differences
```
licehammer: It also changes ownership.
Also, it would be good if you can describe this in detail. This is a significant change in the service behavior, and we should explain it properly.

PR #389, opened 2018-12-07
Fix catch_error exit in safeq

balcirakpeter:
- problem: When Ldifdiff failed, safeq script did not terminate as it supposed to.
           Reason for this behaviour was pipe, which was used after Ldifdif script.
           Instead of terminating the whole script, exit status was captured by the pipe.
- change:  Ldifdiff result is send to temporary file.
	   Next command reads data from this file instead of getting them trough pipe.
- result:  When Ldifdiff fails, the whole script terminate.
	   If the Ldifdiff pass, result is stored in temporary file, which is read by next command.

### DIFF
--- a/send/safeq
+++ b/send/safeq
@@ -102,7 +102,8 @@ OLD_FILE="$DESTINATION_DIR/safeq-content.ldif"
 SINFILE=`mktemp --tmpdir=/tmp sorted-safeq-XXXXXX.ldif`
 S_OLD_FILE=`mktemp --tmpdir=/tmp sorted-safeq-old-XXXXXXX.ldif`
 
-FINAL=`mktemp safeq-final-XXXXXX.ldif`
+LDIF=`mktemp --tmpdir=/tmp safeq-ldif-XXXXXXX.ldif`
+FINAL=`mktemp --tmpdir=/tmp safeq-final-XXXXXX.ldif`
 
 if test -s "$OLD_FILE"; then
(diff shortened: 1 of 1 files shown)


zlamalp:
@licehammer can you please look at this. It helps safeq to prevent sending out malformed LDIF if sorting and diffing fails for some reason.

send/safeq
```diff
@@ -112,7 +113,8 @@ if test -s "$OLD_FILE"; then
 	catch_error E_LDIFSORT_FILE $LDIFSORT -k dn $INFILE >$SINFILE
 
 	# DIFF LDIFs
-	catch_error E_LDIFDIFF_FILE $LDIFDIFF -k dn $SINFILE $S_OLD_FILE | sed '1,1{/^$/d; }; /^[^ ].*/N; s/\n //g' | $SAFEQ_LDIF_SORT > "$FINAL"
+	catch_error E_LDIFDIFF_FILE $LDIFDIFF -k dn $SINFILE $S_OLD_FILE > $LDIF
+	sed '1,1{/^$/d; }; /^[^ ].*/N; s/\n //g' $LDIF > $SAFEQ_LDIF_SORT > "$FINAL"
```
licehammer:
```suggestion
	sed '1,1{/^$/d; }; /^[^ ].*/N; s/\n //g' $LDIF | $SAFEQ_LDIF_SORT > "$FINAL"
```

PR #390, opened 2019-01-07
Relation type updated in vsup_ifis service

zlamalp:
- We now use "EXT" for external users in vsup_ifis service. It`s now
  the same as in other services. We now also have only one type of
  external users -> teaching staff (constant used for DB inserts).


stavamichal:
This logic will be skipped. There is no interest in it.

PR #403, opened 2019-04-02
New version of service fs_home

stavamichal:
- there is major change in the processing data and also in the protocol
 version
 - processing data is now divided into two separate files, first one
 with data needed for creating home directories for users, second one
 with data to set quotas on all defined volumes
 - administrator of the facility can now choose which attributes for
 quotas will be used by setting attribute 'readyForNewQuotas' to 'true'
 if new attributes should be used or to 'false' if old attributes should
 be used. Format of output data file is the same for both logic
 - slave script was modified to respect these changes so first all home
 directories are created and then all quotas are changed
 - IMPORTANT: because the protocol version has been changed, new package
 with service fs_home need to be updated. Also pre|post scripts and mid
 hooks should be check if they still work with new modifications in the
 slave script.

### DIFF
--- a/gen/fs_home
+++ b/gen/fs_home
@@ -7,8 +7,8 @@ use File::Basename;
 no if $] >= 5.017011, warnings => 'experimental::smartmatch';
 
 our $SERVICE_NAME = basename($0);
-our $PROTOCOL_VERSION = "3.6.0";
-my $SCRIPT_VERSION = "3.0.9";
+our $PROTOCOL_VERSION = "3.7.0";
+my $SCRIPT_VERSION = "3.0.10";
 
 sub mergeQuotas {
 	my ($sumQuota, $resourceQuota, $memberQuota) = @_;
@@ -44,6 +44,8 @@ my $data = perunServicesInit::getDataWithGroups;
 
 #Constants
 our $A_USER_LOGIN;                 *A_USER_LOGIN =                   \'urn:perun:user_facility:attribute-def:virt:login';
+our $A_UF_DATA_QUOTAS;             *A_UF_DATA_QUOTAS =               \'urn:perun:user_facility:attribute-def:virt:dataQuotas';
+our $A_UF_FILE_QUOTAS;             *A_UF_FILE_QUOTAS =               \'urn:perun:user_facility:attribute-def:virt:fileQuotas';
 our $A_USER_STATUS;                *A_USER_STATUS =                  \'urn:perun:member:attribute-def:core:status';
 our $A_HOME_MOUNTPOINT;            *A_HOME_MOUNTPOINT =              \'urn:perun:resource:attribute-def:def:fsHomeMountPoint';
 our $A_R_VOLUME;                   *A_R_VOLUME =                     \'urn:perun:resource:attribute-def:def:fsVolume';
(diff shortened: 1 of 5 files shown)


stavamichal:
There is still missing service fs_home_mu, which will be added after approving of this new concept of the service.

gen/fs_home
```diff
@@ -147,16 +153,117 @@ foreach my $login (sort keys %$memberAttributesByLoginAndMount) {
...
+			my %memberAttributes = attributesToHash $mData->getAttributes;
+			my $uid = $memberAttributes{$A_UID};
+
+			unless($dataForUserQuotasByUID->{$uid}) {
+				#First go through data quota map and read file quota if exists or set unlimited
+				foreach my $volume (keys %{$memberAttributes{$A_UF_DATA_QUOTAS}}) {
+					my $dataQuota = $memberAttributes{$A_UF_DATA_QUOTAS}{$volume};
+
+					my $softDataQuota = $dataQuota;
+					$softDataQuota =~ s/:.*$//g;
```
licehammer:
```suggestion
					$softDataQuota =~ s/:.*$//;
```

gen/fs_home
```diff
@@ -147,16 +153,117 @@ foreach my $login (sort keys %$memberAttributesByLoginAndMount) {
...
+					my $hardDataQuota = $dataQuota;
+					$hardDataQuota =~ s/^.*://g;
```
licehammer:
```suggestion
					$hardDataQuota =~ s/^.*://;
```

gen/fs_home
```diff
@@ -62,22 +64,34 @@ our $A_F_QUOTAENABLED;             *A_F_QUOTAENABLED =               \'urn:perun
...
 our $A_R_VO_NAME;                  *A_R_VO_NAME =                    \'urn:perun:resource:attribute-def:virt:voShortName';
+our $A_F_NEW_QUOTAS_READY;         *A_F_NEW_QUOTAS_READY =           \'urn:perun:facility:attribute-def:def:readyForNewQuotas';
 
 our $STRUC_GROUPS;   *STRUC_GROUPS =  \"groups";
 
+#headers
+my $dataQuotaHeader = "dataQuota";
+my $dataLimitHeader = "dataLimit";                                                                                                      
+my $fileQuotaHeader = "fileQuota";
+my $fileLimitHeader = "fileLimit";
```
licehammer: Constants are usually written in uppercase.

gen/fs_home
```diff
@@ -147,16 +153,117 @@ foreach my $login (sort keys %$memberAttributesByLoginAndMount) {
...
+						$dataForUserQuotasByUID->{$uid}->{$volume}->{$dataLimitHeader} = 0;
+						my $fileQuota = $memberAttributes{$A_UF_FILE_QUOTAS}{$volume};
+						my $softFileQuota = $fileQuota;
+						$softFileQuota =~ s/:.*$//g;
+						my $hardFileQuota = $fileQuota;
+						$hardFileQuota =~ s/^.*://g;
+						$dataForUserQuotasByUID->{$uid}->{$volume}->{$fileQuotaHeader} = $softFileQuota;
+						$dataForUserQuotasByUID->{$uid}->{$volume}->{$fileLimitHeader} = $hardFileQuota;
+					}
+				}
```
licehammer: Lines 205-218 are just repetition of lines 190-197. It's very confusing that you are processing files quota on two places - first during processing of files quota and then once again to add files quota for volumes which were not processed before.
I'd suggest to write this differently.

gen/fs_home
stavamichal: Changed.

gen/fs_home
stavamichal: Fixed.

gen/fs_home
stavamichal: Fixed.

gen/fs_home
stavamichal: You are right. I just copied it from fs_quotas, but it is confusing. Now after rework it should be less confusing I hope.

gen/fs_home
```diff
@@ -147,16 +153,108 @@ foreach my $login (sort keys %$memberAttributesByLoginAndMount) {
...
+					my $dataQuota = $memberAttributes{$A_UF_DATA_QUOTAS}{$volume};
+
+					my $softDataQuota = $dataQuota;
+					$softDataQuota =~ s/:.*$//;
+					my $hardDataQuota = $dataQuota;
+					$hardDataQuota =~ s/^.*://;
+					$dataForUserQuotasByUID->{$uid}->{$volume}->{$DATA_QUOTA_HEADER} = quotaToKb $softDataQuota;
+					$dataForUserQuotasByUID->{$uid}->{$volume}->{$DATA_LIMIT_HEADER} = quotaToKb $hardDataQuota;
+					$dataForUserQuotasByUID->{$uid}->{$volume}->{$FILE_QUOTA_HEADER} = 0;
+					$dataForUserQuotasByUID->{$uid}->{$volume}->{$FILE_LIMIT_HEADER} = 0;
```
licehammer: This works, but it deserves at least a comment in code. If a user will be assigned to 2 or more resources, you will first set data quota to correct value and file quota to zero. Then you will set file quota to correct number. After that you will start to process second resource, so you set again data quota to correct number (which is OK), but you also reset file quota to zero. Then you move to next part when you set file quota back to correct number.

gen/fs_home
licehammer: It's way better :+1:

gen/fs_home
stavamichal: I don't think so. There is this "if clause:
`unless($dataForUserQuotasByUID->{$uid})`
which will prevent reading of these user_facility attributes more than once per facility.

licehammer: Good point. This section processes only user-facility attributes, therefore it's enough for it to run once for each user.

PR #408, opened 2019-04-25
`process-voms`: Sort user records by CA

sustr4:
- Matching DNs issued by different CAs will always come in the same order, probably oldest first.
- This is to fix flapping registrations wherein the VOMS record for a given user is deleted and immediately recreated if `skip-CA-check` is `true` and CAs come in different order from the parsed XML.


sustr4:
Pending functionality test but I'm confident enough to open the PR before the test runs.

PR #412, opened 2019-06-26
Grace period behaviour added to ad_group_mu_ucn

balcirakpeter:
Gen: adGracePeriod resource attribute added which will hold grace period
     time in days for the given resource. If the attribute is not set
     for the resource, value will be set to 0.
Send: Members which are in grace period for some group will be stored in a cache file
      which holds this information. When a group is removed all its
      member records from the file are removed.
      Members updating for group is divided into two parts: adding and
      removing. Adding works as usual and remowing removes only those
      members whose grace preiod expired or the grace period of the
      group is set to 0.

### DIFF
--- a/send/ad_group_mu_ucn
+++ b/send/ad_group_mu_ucn
@@ -1,4 +1,7 @@
 #!/usr/bin/perl
+
+my @today = Today();
+
 use strict;
 use warnings;
 no if $] >= 5.017011, warnings => 'experimental::smartmatch';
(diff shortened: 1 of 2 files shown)


send/ad_group_mu_ucn
```diff
@@ -30,6 +43,13 @@ chomp($facility_name);
 my $service_files_base_dir="../gen/spool";
 my $service_files_dir="$service_files_base_dir/$facility_name/$service_name";
 
+# files used for computing a grace period
+my $MEMBER_STATE_FILE ="spool/$facility_name/ad_group_mu_ucn/memberState";
```
zlamalp: I don't recommend storing 'last state' in the ``gen`` folder. I believe it can be wiped out by e.g. manually started gen or you lost it by renaming facility in Perun. Other services store such files in the ``send`` folder, since it is where the send script is located too.
In CRM service for ceitec we also store previous states in the ``logs/`` subfolder of the send. Its not the best, but reasonable solution if you would need to log/trace previous states and compare them.

PR #414, opened 2019-07-03
Replace SUSPENDED state for new suspension logic for member

stavamichal:
- when member is VALID + newly suspended = SUSPENDED
     - when member is EXPIRED + newly suspended = SUSPENDED
     - replace where needed
     - remove where not needed any more
    
    DEPLOYMENT:
     - virtual attribute need to be added to every affected service manualy
     for every instance where services exists

### DIFF
--- a/gen/bbmri_collections
+++ b/gen/bbmri_collections
@@ -28,6 +28,7 @@ sub processMemberships;
 #Constants
 our $A_USER_ID;                   *A_USER_ID =               \'urn:perun:user:attribute-def:core:id';
 our $A_USER_STATUS;               *A_USER_STATUS =           \'urn:perun:member:attribute-def:core:status';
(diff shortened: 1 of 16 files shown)


stavamichal:
Need to be merged together with https://github.com/CESNET/perun/pull/2288

zlamalp:
Core implementation was merged as CESNET/perun#2848.

zlamalp (changes requested):
Changes are OK, but there are many other services, which handle members status and we need to update them too or decide, that its not necessary. Also it should be re-based.

stavamichal:
It should be rebased and ready for review and merging.

stavamichal:
ad other services: We don't need to work with services that are using only VALID users. These services would work the same way for all situations except SUSPENDED users (which are now VALID + suspended). We should talk with administrators of these services to get information if they want to skip suspended users no matter the status in the VO.

zlamalp (approved):

PR #419, opened 2019-08-13
Generic_send supports propagation to Windows

balcirakpeter:
Two destination types were added to generic_send: user@host-windows and
host-windows-proxy.
user@host-windows will be used as a direct connection to windows server
2019 through ssh. Tar file, which will be sent, is encoded to base64 in
generic_send.
host-windows-proxy will be used as a connection to windows proxy server
(solution for older servers). Tar file, which will be sent, is also encoded to base64 in
generic_send, but in its end is added destination server. Name of the
proxy server has to be defined in the file
/etc/perun/services/generic_send/generic_send.conf in the variable
WINDOWS_PROXY.

### DIFF
--- a/send/generic_send
+++ b/send/generic_send
@@ -22,6 +22,19 @@ DESTINATION_TYPE_EMAIL="email"
 DESTINATION_TYPE_HOST="host"
 DESTINATION_TYPE_USER_HOST="user@host"
 DESTINATION_TYPE_USER_HOST_PORT="user@host:port"
+DESTINATION_TYPE_USER_HOST_WINDOWS="user@host-windows"
+DESTINATION_TYPE_HOST_WINDOWS_PROXY="host-windows-proxy"
+
+#This function converts stdin based on DESTINATION_TYPE variable
+function destination_type_transformation {
+	if [ "$DESTINATION_TYPE" = "$DESTINATION_TYPE_HOST_WINDOWS_PROXY" ]; then
+		base64 | sed -e "\$s/\$/ $DESTINATION/g" #converts stdin to base64 and append single space and "$DESTINATION" at the end of it
+	elif [ "$DESTINATION_TYPE" = "$DESTINATION_TYPE_USER_HOST_WINDOWS" ]; then
+		base64 #converts stdin to base64
+	else
+		cat #just prints stdin to stdout for other destination types
+	fi
+}
 
 #if there is no destination type, use default 'host'
 if [ -z "$DESTINATION_TYPE" ]; then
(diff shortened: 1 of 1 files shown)


send/generic_send
```diff
@@ -22,6 +22,19 @@ DESTINATION_TYPE_EMAIL="email"
...
 DESTINATION_TYPE_USER_HOST="user@host"
 DESTINATION_TYPE_USER_HOST_PORT="user@host:port"
+DESTINATION_TYPE_USER_HOST_WINDOWS="user@host-windows"
+DESTINATION_TYPE_HOST_WINDOWS_PROXY="host-windows-proxy"
+
+#This function runs "$FILTER" as a command.
+#If "$FILTER" equals "$DESTINATION",
+#the function runs base64 command and adds the "$DESTINATION" to the end of its output.
+function filter {
+	if [ "$FILTER" = "$DESTINATION" ]; then
```
licehammer: It's functional, but the code is not nice. 
I would prefer to store only constants to $FILTER (e.g. "none", "base64", "host-windows-proxy") and then modify the result of the filter function based on that.

send/generic_send
```diff
@@ -22,6 +22,19 @@ DESTINATION_TYPE_EMAIL="email"
...
+		base64 | sed -e "\$s/\$/ $DESTINATION/g"
```
licehammer:
```suggestion
		base64 | sed -e "\$s/\$/ $DESTINATION/g" #converts stdin to base64 and append single space and "$DESTINATION" at the end of it
```

send/generic_send
```diff
@@ -84,6 +105,16 @@ case $DESTINATION_TYPE in
 		HOSTNAME="$DESTINATION"
 		HOST="$DESTINATION"
 		;;
+	${DESTINATION_TYPE_USER_HOST_WINDOWS})
+		HOST=`echo $DESTINATION | sed -e 's/:.*//'`
+		HOSTNAME=`echo $HOST | sed -e 's/^.*@//'`
+		FILTER="base64"
+		;;
+	${DESTINATION_TYPE_HOST_WINDOWS_PROXY})
+		if [ -z "${WINDOWS_PROXY}" ]; then echo 'Windows proxy name is not defined in /etc/perun/services/generic_send/generic_send.conf' >&2 ; exit 1; fi
```
licehammer: Put the variable name to the error message. It will be easier to debug.

Also, I would mentioned that the variable is **usually** defined in /etc/perun/...

send/generic_send
balcirakpeter: I have completely erased FILTER variable. The filter function now depends on DESTINATION_TYPE. It seemed to me like the best solution but if it is not ok I will change it the way you wanted.

send/generic_send
balcirakpeter: Added

send/generic_send
balcirakpeter: Fixed

balcirakpeter:
I've changed the function name to destination_type_transformation.

PR #422, opened 2019-08-27
Remove prefix from attribute shibbolethExtSources

stavamichal:
- in attribute shibbolethExtSources there should be prefix for every
 key, because some of them can be same with different logins (values),
 for this reason, we need to strip this prefix before saving it


stavamichal:
Can't be merged without counterpart in perun repo: https://github.com/CESNET/perun/pull/2382

PR #423, opened 2019-08-29
User certificates in ad_mu

balcirakpeter:
- Added userCertificates attribute to ad_mu gen and send scripts.
  Each certificate in pem format is converted to der format and
  encoded to base64, so LDIF can process it in the send script.

### DIFF
--- a/gen/ad_mu
+++ b/gen/ad_mu
@@ -5,6 +5,8 @@ use warnings;
 use perunServicesInit;
 use perunServicesUtils;
 use Time::Piece;
+use MIME::Base64;
+use Encode;
 
 sub processMembers;
 sub processExtendedLicenses;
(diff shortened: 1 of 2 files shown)


balcirakpeter:
Gen script was tested and it works correctly. Send script was not tested because of the Cyclop testing.

gen/ad_mu
```diff
@@ -529,6 +534,11 @@ for my $login (@logins) {
 
 	print FILE "userAccountControl: $facilityAttributes{$A_F_UAC}\n";
 
+	# Convert each user certificate in pem format to der format, encode it to base64 and print it to the file
+	foreach my $certificate_pem (@{$users->{$login}->{$A_U_CERTIFICATES}}) {
+		print FILE "userCertificate:: ", encode_base64(Encode::encode_utf8($certificate_pem), ''), "\n";
+        }
```
stavamichal, line 540: Wrong indentation.

PR #428, opened 2019-09-25
Process HTTP return codes properly in generic send script

stavamichal:
- if the destination is the URL type, we need to process the return
 code from HTTP server properly and exit with it if it is different from
 200 (OK)
 - save standard output to the temporary file, http error to variable
 and manage proper return code and text for such situation
 - in curl use '-s' to remove progress bars if file is transfered
 - in curl use '-w {http_code}' to redirect HTTP_CODE on standard output
 - in curl use '-o $TMPFILE' to redirect standard output of http server
 to temporary file
 - remove '-f' from curl to prevent masking http server errors

### DIFF
--- a/send/generic_send
+++ b/send/generic_send
@@ -49,7 +49,13 @@ else
 	if [ -r "${PERUN_CERT}" -a -r "${PERUN_KEY}" ]; then
 		PERUN_CERT_SETTING="--cert ${PERUN_CERT} --key ${PERUN_KEY}"
 	fi
-	TRANSPORT_COMMAND="curl ${PERUN_CERT_SETTING} -i -H Content-Type:application/x-tar -f -X PUT --data-binary @- "
+	TMPFILE=`mktemp /tmp/perun-generic-send-script.XXXXXX`
+	if [ $? -ne 0 ]; then
+		echo "Can't create TMPFILE /tmp/perun-generic-send-script.XXXXXX" >&2
+		exit 255
+	fi
+	trap 'rm -r -f "$TMPFILE"' EXIT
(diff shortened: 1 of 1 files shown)


send/generic_send
```diff
@@ -49,7 +49,8 @@ else
 	if [ -r "${PERUN_CERT}" -a -r "${PERUN_KEY}" ]; then
 		PERUN_CERT_SETTING="--cert ${PERUN_CERT} --key ${PERUN_KEY}"
 	fi
-	TRANSPORT_COMMAND="curl ${PERUN_CERT_SETTING} -i -H Content-Type:application/x-tar -f -X PUT --data-binary @- "
+	TMPFILE=`mktemp /tmp/perun-generic-send-script.XXXXXX`
```
licehammer: It's rare but still possible for mktemp command to fail. You should add check for this type of error.

stavamichal: Should be fixed now.

stavamichal:
Trap for TMPFILE was added too.

send/generic_send
```diff
@@ -168,6 +174,19 @@ ERR_CODE=$?
...
+		if [ 200 -ne "$STDOUT" ]; then
+			ERR_CODE=$STDOUT
+			cat "$TMPFILE" >&2
+		else
+			ERR_CODE=0
+			cat "$TMPFILE"
+		fi
+	else
+		#if this is not url destination, than we need to print stdout from slave script
+		echo "$STDOUT"
```
balcirakpeter: I think that after the changes which were made in this pull request, this echo will print http code instead of stdout from the slave script (which is in tmpfile now). So, as a result, the slave script message will not be printed at all if the destination type is not URL.

PR #431, opened 2019-10-17
Fix curl timeout in send script

stavamichal:
- when curl ends with timeout (and return code 28), it will return
 http-code "000". We need to process this code and respond with
 proper message and error. For this situation the script will end with
 the same return code as curl (28)

### DIFF
--- a/send/generic_send
+++ b/send/generic_send
@@ -55,7 +55,7 @@ else
 		exit 255
 	fi
 	trap 'rm -r -f "$TMPFILE"' EXIT
-	TRANSPORT_COMMAND="curl ${PERUN_CERT_SETTING} -i -H Content-Type:application/x-tar -w %{http_code} --silent -o $TMPFILE -X PUT --data-binary @- "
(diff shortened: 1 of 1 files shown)


send/generic_send
```diff
@@ -176,7 +176,10 @@ if [ $ERR_CODE -eq 124 ]; then
 else
 	if [ "$DESTINATION_TYPE" = "$DESTINATION_TYPE_URL" ]; then
 		#the return code from curl is saved in STDOUT
-		if [ 200 -ne "$STDOUT" ]; then
+		if [ "000" = "$STDOUT" ]; then
+			ERR_CODE=28
+			echo "Curl has ended with error code 28 (timeout)!" >&2
```
licehammer: Unfortunately, you cannot be sure that "000" in $STDOUT means that curl ended with error code 28. "000" means basically, any client-side error.

Best solution would be to check $ERR_CORE and only if it's zero then check $STDOUT. 

Also, I would recommend to add `--show-error` to curl options which should produce error message from curl in case any error will occur.

stavamichal: Changed to cover all possibilities.

stavamichal:
Please @licehammer check this PR again. Thank you.

PR #437, opened 2019-11-06
Detect duplicities in drupal elixir service

stavamichal:
- when generating data for drupal_elixir service, we need to detect
   situations, where two or more users with different logins has the
   same email (in lower-case format). In such situation, we want to find
   all duplicities, print them as an error and end gen script with an error.

### DIFF
--- a/gen/drupal_elixir
+++ b/gen/drupal_elixir
@@ -52,6 +52,8 @@ our $g_type = {};
 our $g_parent_id = {};
 
 our $membershipStruc = {};
+our $usedEmailsStruc = {};
+our $errorInDuplicity;
 
 my $fileUsers = $DIRECTORY . "users.csv";
 my $fileGroups = $DIRECTORY . "groups.csv";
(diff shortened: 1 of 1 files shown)


gen/drupal_elixir
```diff
@@ -104,6 +106,9 @@ foreach my $uid (sort keys %$membershipStruc) {
 }
 close (FILE_MEMBERSHIPS) or die "Cannot close $fileMemberships: $! \n";
 
+#duplicity in email is not allowed state
+if($errorInDuplicity) { die "ERROR: There are some duplicities in emails:\n" . $errorInDuplicity; }
```
zlamalp: Unless you need printed data in files for some check, we can probably fail this right away on line 71, since check on duplicity is already done when processing groups->users->memberships.

balcirakpeter: Or it can die right in the processUsers sub when the duplicity occurs. It is not necessary to continue to another user, because the script will die anyway.

stavamichal: Moved.

PR #452, opened 2020-03-24
LDAP request changes

balcirakpeter:
- Added new functionality resolve_domain_controlers to ADConnector.pm,
  which resolve all available domain controlers from a hostname.
  Added also method ldap_connect_multiple_options which try to connect
  to the first available controler from the list of controlers.
- Updated services for this functionality used in MUNI AD.
  ad_ceitec is set up manually and vsup services cannot be updated yet because,
  the NET::DNS::Resolver perl module need to be installed on the server first.

### DIFF
--- a/send/ADConnector.pm
+++ b/send/ADConnector.pm
@@ -1,7 +1,7 @@
 package ADConnector;
 use Exporter 'import';
 @ISA = ('Exporter');
-@EXPORT = qw(init_config resolve_pdc ldap_connect ldap_bind ldap_unbind ldap_log load_perun load_ad load_group_members compare_entry enable_uac disable_uac is_uac_enabled);
+@EXPORT = qw(init_config resolve_domain_controlers ldap_connect_multiple_options resolve_pdc ldap_connect ldap_bind ldap_unbind ldap_log load_perun load_ad load_group_members compare_entry enable_uac disable_uac is_uac_enabled);
 
 use strict;
 use warnings;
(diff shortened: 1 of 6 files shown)


balcirakpeter:
This functionality will be implemented also for VSUP services but it needs to be discussed with them firstly.

send/ADConnector.pm
```diff
@@ -118,6 +127,59 @@ sub init_config($) {
...
+		die "[AD] Ldap location in wrong format. Expected: protocol://host:port";
+	}
+	my $resolver = Net::DNS::Resolver->new;
+	my $query = $resolver->search($host);
+	my @controlers = ();
+
+	if ($query) {
+  		foreach ($query->answer) {
+    		next unless $_->type eq "A";
+			push(@controlers, $protocol . $_->address . ":" . $port);
```
licehammer: Indent ;-)

send/ADConnector.pm
```diff
@@ -118,6 +127,59 @@ sub init_config($) {
...
+  		}
+	} else {
+		my $error = $resolver->errorstring();
+		ldap_log('ad_connection', "[AD] No answer was found for $host. DNS query returned error message: $error");
+        print STDERR "[AD] No answer was found for $host. DNS query returned error message: $error";
+        die "[AD] No answer was found for $host. DNS query returned error message: $error";
```
licehammer: Indent ;-)

Also, print to STDERR is not necessary because that's already covered by die.

send/ADConnector.pm
```diff
@@ -118,6 +127,59 @@ sub init_config($) {
 
 }
 
+#
+# Resolve available domain controlers from a URL
+#
+sub resolve_domain_controlers($) {
+	my $ldap_location = shift;
+	my ($protocol, $host, $port);
+	if($ldap_location =~ /^([^\/]+:\/\/)?([^:]+):([0-9]+)$/) {
```
licehammer: Usually you are using a space between "if" and "(". It's good to be consistent.

balcirakpeter: Fixed

send/ADConnector.pm
balcirakpeter: Fixed

send/ADConnector.pm
balcirakpeter: Fixed

PR #453, opened 2020-04-02
RemoveGrace period in ad_group_mu_ucn

balcirakpeter:
- Grace period was removed from ad_group_mu_ucn logic.
- Gen was also reworked, so it skips expired members.
  It is done by calling getHierarchicalData with
  filter expired members oprtion set to true.
- This changes were made because the grace period is moved to perun.


balcirakpeter:
PR moze byt mergnuty az po tom ako budu grace periody nachystane na strane peruna.
Zaroven je este nutne rucne otestovanie zmien!

balcirakpeter:
Gen script was tested on idm-test and it works. Send part was not tested but it is just old version of ad_group_mu_ucn script which were used before grace periods.

PR #456, opened 2020-05-19
Changed logging for ad_group_mu_ucn

balcirakpeter:
- send script ad_group_mu_ucn used to print out the whole ldif while
  updating users in AD.
- Now only users whose send script tries to add or delete are printed
  into the log file.


balcirakpeter:
We decided that the whole script needs to be rewritten. Therefore, this PR will be closed.

PR #459, opened 2020-06-03
Removed unimportant logs for service k5login

stavamichal:
- skip information about not changed k5login files. This
 information is not important for administrators of a service

### DIFF
--- a/slave/process-k5login/changelog
+++ b/slave/process-k5login/changelog
@@ -1,3 +1,10 @@
+perun-slave-process-k5login (3.1.6) stable; urgency=low
+
+  * Skip information about not changed k5login files. This information is
+    not important for administrators of a service.
+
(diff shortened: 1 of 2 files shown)


slave/process-k5login/changelog
```diff
@@ -1,3 +1,10 @@
+perun-slave-process-k5login (3.1.6) stable; urgency=low
+
+  * Skip information about not changed k5login files. This information is
+    not important for administrators of a service.
+
+ -- Michal Stava <stavamichal@gmail.com> Wed, 03 Jun 2020 14:57:00 +0100
```
zlamalp, line 6: Should be +0200, since we currently have CEST

stavamichal: Fixed.

PR #465, opened 2020-06-23
ad_mu process members in chunks

balcirakpeter:
- updating members in ordinary groups and groups for additional licenses was reworked.
      Now it uses subroutine update_group_membership which handles categorizing members,
      calling new add and remove subtroutines and updating counters.
    - subroutines add_members_to_entry and remove_members_from_entry just
      performe these operations in chunks of max 4999 members, logs results
      for each chunk and return 0 if there was no error and 1 if at least one
      chunk could not been updated.
    - updating members in main licence groups had to be rewritten
      differently. It is because for the all main licence groups need to be
      firstly performed adding operation and after that can be performed removing operation.
      It is because these groups are hooked to some external process in AD
      which can be run in the middle of the update.
    - In order to keep this behaviour, updating main licence groups is using
      the same add and remove subroutins as before, but these subroutins were changed.
      They now use subroutins which works with chunks. They then return messages
      based on the result of the chunk subrtoutins. "unchanged" means that nothing changed.
      "errors" mean that update process was called but there was some errors.
      "changed" means that update was performed without any problems.
      According to these messages are updated counters.
    - All these changes were made because LDAP could not perform update with
      more than 5000 changes.

### DIFF
--- a/send/ad_mu
+++ b/send/ad_mu
@@ -324,6 +324,9 @@ my $USER_STATE_GP_NOTIFICATION_SENT_REMINDER = "2";
 my $GRACE_PERIOD_EXPIRATION = 301; #days
 my $GRACE_PERIOD_EXPIRATION_NOTIFICATION_DELTA_DAYS = 14; #days
 
+my $SUCCESS = 1;
+my $FAIL = 0;
+
 
 my $facility_name = $ARGV[0];
 chomp($facility_name);
@@ -393,6 +396,11 @@ my $studentDN = "CN=O365Lic_Student_group.muni.cz,OU=licenses," . $base_dn_group
 my $student2DN = "CN=O365Lic_Student2_group.muni.cz,OU=licenses," . $base_dn_groups;
 my $alumniDN = "CN=O365Lic_Alumni_group.muni.cz,OU=licenses," . $base_dn_groups;
 
+# operations results
+my $RESULT_ERRORS = "errors";
+my $RESULT_CHANGED = "changed";
+my $RESULT_UNCHANGED = "unchanged";
+
 # log counters
 my $counter_add = 0;
 my $counter_updated = 0;
@@ -403,6 +411,7 @@ my $counter_fail_ous = 0;
 my $counter_group_add = 0;
 my $counter_group_remove = 0;
 my $counter_group_updated = 0;
+my $counter_group_updated_with_errors = 0;
 my $counter_group_failed = 0;
 
 # load all data
@@ -465,7 +474,8 @@ ldap_log($service_name, "Disabled: " . $counter_disabled . " entries.");
 ldap_log($service_name, "Failed: " . $counter_fail. " entries.");
 ldap_log($service_name, "Group added: " . $counter_group_add . " entries.");
 ldap_log($service_name, "Group removed: " . $counter_group_remove . " entries.");
-ldap_log($service_name, "Group updated: " . $counter_group_updated . " entries.");
+ldap_log($service_name, "Group updated without errors: " . $counter_group_updated . " entries.");
+ldap_log($service_name, "Group updated with errors: " . $counter_group_updated_with_errors . " entries.");
 ldap_log($service_name, "Group failed: " . $counter_group_failed . " entries.");
 
 # print results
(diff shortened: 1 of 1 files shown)


send/ad_mu
```diff
@@ -1025,22 +984,32 @@ saveStoredUserStates;
...
+			$counter_group_updated++;
+		}
+	}
 
 	# process specific license groups
 	foreach my $group_dn (sort keys %{$perun_state}) {
 		unless (($group_dn eq $employeeDN) or ($group_dn eq $studentDN) or ($group_dn eq $alumniDN) or ($group_dn eq $student2DN)) {
-			add_to_license_group($ad_entries_group_map{$group_dn}, $ad_state->{$group_dn}, $perun_state->{$group_dn});
-			remove_from_license_group($ad_entries_group_map{$group_dn}, $ad_state->{$group_dn}, $perun_state->{$group_dn});
+			update_group_membership($ad_entries_group_map{$group_dn}, $ad_state->{$group_dn}, $perun_state->{$group_dn});
```
zlamalp, line 1012: Former implementation of ``add_to_license_group`` and ``remove_from_license_group`` exits the script on LDAP change failure. By using new implementation, it continues the script.

It might not be desired, since failure in adding of member and success of removing will cause inconsistency for the mentioned "process inside LDAP" and it might result in storing this state in local caches and sending unnecessary notifications to the user.

Also ``update_group_membership`` here uses generic methods to add/remove members, so it probably won't be subject of possible fixes for managing license groups.

send/ad_mu
```diff
@@ -477,7 +479,8 @@ print "OU added: " . $counter_add_ous . " entries.\n";
 print "OU failed: " . $counter_fail_ous . " entries.\n";
 print "Group added: " . $counter_group_add . " entries.\n";
 print "Group removed: " . $counter_group_remove . " entries.\n";
-print "Group updated: " . $counter_group_updated . " entries.\n";
+print "Group updated without errors: " . $counter_group_updated . " entries.\n";
+print "Group updated with errors: " . $counter_group_updated_with_errors . " entries.\n";
```
zlamalp, line 483: You should probably add ``$counter_group_updated_with_errors`` in the condition, which exits the script with code 1 to re-schedule the propagation (few lines below).

balcirakpeter: Added

send/ad_mu
balcirakpeter: I've added check after processing the main license groups, which ends the script if one of them fails.

send/ad_mu
```diff
@@ -1025,22 +984,38 @@ saveStoredUserStates;
...
+	$add_result{$alumniDN} = add_to_license_group($ad_entries_group_map{$alumniDN}, $ad_state->{$alumniDN}, $alumni);
+	$add_result{$studentDN} = add_to_license_group($ad_entries_group_map{$studentDN}, $ad_state->{$studentDN}, \%studentsFirstPart);
+	$add_result{$student2DN} = add_to_license_group($ad_entries_group_map{$student2DN}, $ad_state->{$student2DN}, \%studentsSecondPart);
 
-	remove_from_license_group($ad_entries_group_map{$employeeDN}, $ad_state->{$employeeDN}, $employees);
-	remove_from_license_group($ad_entries_group_map{$alumniDN}, $ad_state->{$alumniDN}, $alumni);
-	remove_from_license_group($ad_entries_group_map{$studentDN}, $ad_state->{$studentDN}, \%studentsFirstPart);
-	remove_from_license_group($ad_entries_group_map{$student2DN}, $ad_state->{$student2DN}, \%studentsSecondPart);
+	foreach ($employeeDN, $alumniDN, $studentDN, $student2DN) {
+		if ($add_result{$_} eq "errors") {
```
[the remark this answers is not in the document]
balcirakpeter: Created

send/ad_mu
```diff
@@ -2013,3 +1952,97 @@ sub haveOnlyAlumniRole {
 	my $role = shift;
 	return $role == $F_ALUMNI;
 }
+
+sub add_members_to_entry {
+
+	my $ad_entry = shift;
+	my $to_be_added = shift;
+	my $return_code = 0;
```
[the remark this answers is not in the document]
balcirakpeter: Fixed

send/ad_mu
```diff
@@ -2013,3 +1957,97 @@ sub haveOnlyAlumniRole {
 	my $role = shift;
 	return $role == $F_ALUMNI;
 }
+
+sub add_members_to_entry {
+
+	my $ad_entry = shift;
+	my $to_be_added = shift;
+	my $return_code = 1;
```
stavamichal, line 1965: This is correct, but I would prefer something like:
```
our $SUCCESS = 1;
our $FAIL = 0;

my $return_code = $SUCCESS;
```
The same in all places where you are using these return states.

balcirakpeter: Done

stavamichal (approved):

PR #466, opened 2020-06-23
Updated both google_groups and vsup_google_groups services

zlamalp:
- Use new jar name: google-group-connector-[version].jar
  for the connector app.
- Fixed all 'shellcheck' warnings in the bash script.

Depends on changes from CESNET/google-group-connector#19. Do not merge/deploy without it.


zlamalp:
Manually deployed and tested on both instances, works as expected. I'm merging it.

PR #469, opened 2020-07-02
Added filling of guest group in ad_user_mu_ucn

balcirakpeter:
- ad_user_mu_ucn was creating and updating guests in ucn.muni.cz.
  Now it is also adding these users to the group HostPerun which is
  located in a separate tree, so the ad_group_mu_ucn cannot empty the
  group.


zlamalp:
This change is no longer required by the requestor. I'm closing this pull-request.

PR #470, opened 2020-07-08
Add support of windows connection for service safeq

stavamichal:
- connection was add based on code from general_send


stavamichal:
I will change beahvior from scratch.

PR #472, opened 2020-07-16
added facility attribute for users DN in AD

rosinec:
- added facility attribute for Users DN in AD


gen/safeq
```diff
@@ -21,6 +21,8 @@ my $DIRECTORY = perunServicesInit::getDirectory;
 my $data = perunServicesInit::getHierarchicalData;
 
 #Constants
+our $A_F_BASE_DN;  *A_F_BASE_DN = \'urn:perun:facility:attribute-def:def:adBaseDN';
```
stavamichal, line 24: Please, use the same indentation like in other cases.

PR #475, opened 2020-08-07
Updated send/BBMRI_* scripts 

dBucik:
- updated the messages to be more descriptive about from where the error
code can actually be
- added handling of return code 0
- added correct exiting
- process JSON files with jq instead of cat and compact the output
before sending it to curl

### DIFF
--- a/send/bbmri_collections
+++ b/send/bbmri_collections
@@ -31,25 +31,40 @@ USER_TRANSPORT_COMMAND="$TRANSPORT_COMMAND -d @- $USERNAME_AND_PASSWORD $DESTINA
 GROUP_TRANSPORT_COMMAND="$TRANSPORT_COMMAND -d @- $USERNAME_AND_PASSWORD $DESTINATION$GROUP_ENDPOINT"
 
 # HTTP Request with timeout for USERS
-cat "$USER_FILE" | timeout -k $TIMEOUT_KILL $TIMEOUT $USER_TRANSPORT_COMMAND
+jq -c '.' "$USER_FILE" | timeout -k $TIMEOUT_KILL $TIMEOUT $USER_TRANSPORT_COMMAND
 # Catch errors
-ERR_CODE=$?
-#Error code 124 means - timed out
-if [ $ERR_CODE -eq 124 ]; then
-	echo "Communication with the peer has timed out with return code: $ERR_CODE (Warning: this error can mask original error 124 from peer!)" >&2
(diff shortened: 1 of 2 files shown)


send/bbmri_collections
```diff
@@ -36,9 +36,9 @@ cat "$USER_FILE" | timeout -k $TIMEOUT_KILL $TIMEOUT $USER_TRANSPORT_COMMAND
 ERR_CODE=$?
 #Error code 124 means - timed out
 if [ $ERR_CODE -eq 124 ]; then
-	echo "Communication with the peer has timed out with return code: $ERR_CODE (Warning: this error can mask original error 124 from peer!)" >&2
```
stavamichal, line 39: The warning should stay at the line with a timeout but the description must be better.
The warning on line 52 should be moved to the line 50 in both scripts.

send/bbmri_collections
```diff
@@ -31,25 +31,40 @@ USER_TRANSPORT_COMMAND="$TRANSPORT_COMMAND -d @- $USERNAME_AND_PASSWORD $DESTINA
...
+        if [ $ERR_CODE -eq 124 ]; then
+                echo "Communication with the peer ends with return code: $ERR_CODE (timeout). Warning: this error can mask original error 124 from the transportation process!" >&2
+        else
+                echo "Communication with the peer ends with return code: $ERR_CODE" >&2
+        fi
+        exit $ERR_CODE
 else
-	echo "Communication with the peer ends with return code: $ERR_CODE" >&2
+        echo "Communication with the peer ends with return code: $RES_CODE"
 fi
```
stavamichal, line 49: Ok, this seems to be a little complicated. What about to change code this way:
```
ERR_CODE=$?
if [ "$ERR_CODE" -eq 0 ]; then
  echo "... file was processed ok ..."
elif [ "$ERR_CODE" -eq 124 ]; then
  echo "...problem with timeout ..."
 exit $ERR_CODE
else
  echo "...any other problem ..."
  exit $ERR_CODE
fi
```
The same for both parts and at the last row "exit $ERR_CODE"
I don't see any reason to use two variables, cause you have only three possible options:

1. users command ended with an error  (exit)
2. groups command ended with an error (users are ok, groups are exiting)
3. both commands are ok (exit 0 at the end of the script)

What do you think? Do I miss something?

zlamalp (changes requested):
Please rebase this pull-request, thank you.

PR #477, opened 2020-08-19
ad_group_mu_ucn reworked updating members

balcirakpeter:
- Members are now updated in chunks as it will be in ad_mu. Also its
  loging is improved so it will be more readable.
- When all PRs will be merged, subs which are used can be extracted to
  AD script so we prevent duplicities.

### DIFF
--- a/send/ad_group_mu_ucn
+++ b/send/ad_group_mu_ucn
@@ -19,10 +19,14 @@ sub process_update;
 my $counter_add = 0;
 my $counter_remove = 0;
 my $counter_update = 0;
+my $counter_updated_with_errors = 0;
 my $counter_fail = 0;
 
 # define service
 my $service_name = "ad_group_mu_ucn";
+
+my $FAIL = 0;
+my $SUCCESS = 1;
 
 # GEN folder location
 my $facility_name = $ARGV[0];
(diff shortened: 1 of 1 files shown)


send/ad_group_mu_ucn
```diff
@@ -80,18 +81,20 @@ ldap_unbind($ldap);
...
 print "Removed: " . $counter_remove . " entries.\n";
-print "Updated: " . $counter_update. " entries.\n";
+print "Updated without errors: " . $counter_update . " entries.\n";
+print "Updated with errors: " . $counter_updated_with_errors . " entries.\n";
 print "Failed: " . $counter_fail. " entries.\n";
 
 $lock->unlock();
 
-if ($counter_fail > 0) { die "Failed to process: " . $counter_fail . " entries.\nSee log at: ~/perun-engine/send/logs/$service_name.log";}
+if ($counter_fail > 0 or $counter_updated_with_errors) { die "Some entries failed to process or some members were not updated.\nSee log at: ~/perun-engine/send/logs/$service_name.log";}
```
stavamichal, line 97: I would change this to two parts:
```
if ($counter_fail > 0) { die "....";}
if ($counter_updated_with_errors > 0) {die "....";}
```
Message will be more accurate and you will have the info in the printed results.

send/ad_group_mu_ucn
```diff
@@ -214,3 +202,97 @@ sub process_update() {
 	}
 
 }
+
+sub add_members_to_entry {
+
+	my $ad_entry = shift;
+	my $to_be_added = shift;
+	my $return_code = 1;
```
stavamichal, line 210: This is correct, but I would prefer something like:
```
our $SUCCESS = 1;
our $FAIL = 0;

my $return_code = $SUCCESS;
```
The same in all places where you are using these return states.

stavamichal (changes requested):

send/ad_group_mu_ucn
balcirakpeter: Done

send/ad_group_mu_ucn
balcirakpeter: Done

stavamichal (approved):

PR #478, opened 2020-09-02
Added grace period for groups in ad_mu

balcirakpeter:
- Groups in ad_mu were not deleted from AD but attribute extensionAttribute1 was
  set to FALSE instead. That caused that further process in O365 removed
  these groups from MUNI internal systems.
- When someone accidentaly removed some group in Perun, then the group was
  deleted by the O365 even that the group still existed in AD. When the
  group was created again in Perun, ad_mu script set extensionAttribute1
  to TRUE (in AD), but the O365 created completely new group with a new
  identifier so the internal MUNI systems did not know that it is the
  same group.
- To prevent these accidental removals, grace period was implemented for
  all groups propagated by the ad_mu script. When a group is deleted in
  Perun, the attribute mentioned above is not set immediately to FALSE.
  Instead of that current date + 30 days is set to the new attribute
  extensionAttribute2, which represents grace period for the given group.
  This attribute is initially undef and when the group is restored in
  Perun, it is again set to undef. When a date is set in this attribute
  and this date is before today (meaning when the scripts run), then the
  extensionAttribute1 is set to FALSE.

### DIFF
--- a/send/ad_mu
+++ b/send/ad_mu
@@ -1,6 +1,7 @@
 #!/usr/bin/perl
 
 my @today = Today();
+my $groups_grace_period_date = Date_to_Text(Add_Delta_Days(@today, 30));  # grace period for groups
 #@today = Add_Delta_Days(@today, 155);  #DEBUG - add more day 
 #@today = (2016, 1, 1); #DEBUG - set fo fixed date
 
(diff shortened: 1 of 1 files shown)


balcirakpeter:
The code was tested on the idm_test instance. The only issue is that when it will be deployed, all groups which are set to "FALSE" right now will have the grace period date set to the future (but the "FALSE" sign will stay as it is now).

stavamichal (approved):

PR #481, opened 2020-09-11
Changes in service drupal_elixir

stavamichal:
- we need to check duplicities on slave part (instead of gen part) to
 be able to process whole script successufuly and send list of
 duplicities as error to Peruna engine

### DIFF
--- a/gen/drupal_elixir
+++ b/gen/drupal_elixir
@@ -9,8 +9,8 @@ use Perun::GroupsAgent;
 use open qw/:std :utf8/;
 
 local $::SERVICE_NAME = "drupal_elixir";
-local $::PROTOCOL_VERSION = "3.0.0";
-my $SCRIPT_VERSION = "3.0.1";
+local $::PROTOCOL_VERSION = "3.1.0";
+my $SCRIPT_VERSION = "3.1.1";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
@@ -46,6 +46,9 @@ our $u_email = {};
 our $u_name = {};
 our $u_eppn = {};
 
+our $duplicityUserStruc = {};
+our $lowerCaseEmailStruc = {};
+
 our $groupStruc = {};
 our $g_name = {};
 our $g_type = {};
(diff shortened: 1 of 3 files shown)


slave/process-drupal-elixir/bin/process-drupal_elixir.sh
```diff
@@ -36,4 +38,10 @@ function process {
 			log_msg I_NOT_CHANGED
 		fi
 	done
+
+	#there are some duplicities, need to end with error
+  if [ -s "$FILE_USERS_DUPLICITIES" ]; then
+		cat "$E_DUPLICITIES" >&2;
+		log_msg E_DUPLICITIES
```
zlamalp, line 45: Shouldn't one of them print content of ``users-duplicities.csv`` ? or do you just want to end up with the error code alone?

gen/drupal_elixir
```diff
@@ -147,19 +154,20 @@ sub processUsers {
...
-		#duplicity of two different users in email is not allowed
-		my $lowerCaseEmail = lc($email);
-		if($usedEmailsStruc->{$lowerCaseEmail}) {
-			$errorInDuplicity .= "Users with logins '$login' and '" . $usedEmailsStruc->{$lowerCaseEmail} . "' have same email '$email'\n";
+	} else{
+    my $lowerCaseEmail = lc($email);
+		#duplicity found, create record of duplicity accounts
+		if(exists $lowerCaseEmailStruc->{$lowerCaseEmail}) {
+      $duplicityUserStruc->{$uid}->{$u_login} = $login;
+			$duplicityUserStruc->{$uid}->{$u_email} = $email;
```
zlamalp, line 162: So, the duplicates are not pushed to the normal file. Regarding the managed service - what if duplicity happens later, when first account already exist. Do we care? Is such account removed/disabled?

slave/process-drupal-elixir/bin/process-drupal_elixir.sh
stavamichal: You are right, there is wrong variable. Fixed.

gen/drupal_elixir
stavamichal: We are just sending file with actual situation without duplicities. There is possibility when one account will be removed and different one will be added. I don't think we can do anything about it.

zlamalp: Ok, I was just thinking aloud. If service didn't specify its preferences for solution, this should be ok.

zlamalp (approved):

balcirakpeter (approved):

PR #484, opened 2020-09-17
updated UCN AD provisioning to match requirements

rosinec:
- Ignoring Guest- and L- accounts,
- Removing membership of unmanaged groups,
- Support multiple OUs per group,
- enabling all accounts to be provisioned to AD,
- filling new attributes for accounts.


stavamichal:
You probably merged all changes to this commit. You should always rebase PR to the actual master. I am afraid this can't be merged in this state. You should separate all logic you added and create a new PR with that.

PR #486, opened 2020-09-30
Suppress unnecessary TAR warnings about timestamps

zlamalp:
- Do not warn about file timestamps in the future
  when unpacking compressed data from Perun in the
  main slave script.
  It causes propagation to stay in WARNING state
  instead of DONE.


zlamalp:
Tested manually, works.

PR #492, opened 2020-10-15
Removed duplicities in ad services

balcirakpeter:
- Methods add_members_to_entry, remove_members_from_entry and
  update_groupMembership were moved to the ADConnector.pm to prevent
  duplicities in ad_mu and ad_group_mu_ucn scripts.
- These methods had to be slightly changed as they cannot update
  counters. Instead of that they return result codes, which are used in
  the destination scripts to update appropriate counters.

### DIFF
--- a/send/ADConnector.pm
+++ b/send/ADConnector.pm
@@ -1,7 +1,7 @@
 package ADConnector;
 use Exporter 'import';
 @ISA = ('Exporter');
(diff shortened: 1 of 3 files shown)


balcirakpeter:
All changes were tested both in the test and production environment.

stavamichal (approved):

zlamalp (approved):

PR #493, opened 2020-10-20
Removed unused variable from the script for scim service

zlamalp:
- There were unused variables, shadowed by local versions
  in subs. They are now removed.
- Removed unused assignment and call to getAgent().

### DIFF
--- a/gen/scim
+++ b/gen/scim
@@ -4,8 +4,6 @@ use strict;
 use warnings;
 use perunServicesInit;
 use perunServicesUtils;
-use Perun::Agent;
-use Perun::GroupsAgent;
 use open qw/:std :utf8/;
 use JSON::XS;
 use utf8;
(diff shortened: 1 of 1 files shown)


balcirakpeter (commented):
I've got a few comments not related to your change:
1. The script is not executable.
2. The script does nothing.
3. There can be removed also unused imports for Agent and GroupsAgent.
4. Also sub declarations at the beginning of the script can be removed.

zlamalp:
> I've got a few comments not related to your change:
> 
>     1. The script is not executable.
Fixed
> 
>     2. I've tested it on devel and it returns empty files (not sure why).
Misconfiguration on devel. Now it works.
> 
>     3. There can be removed also unused imports for `Agent` and `GroupsAgent`.
Fixed
> 
>     4. Also sub declarations at the beginning of the script can be removed.
They can't if we won't rearrange all the code. Subs must be declared before first usage or script will complain. With this forward declaration we can use them anywhere and main code can be at the script beginning. I would prefer to keep it this way, since its like this in all our scripts.
> 
>     5. Result in `groupMembersLogins` is not used.
Assignment removed.

balcirakpeter:
> They can't if we won't rearrange all the code. Subs must be declared before first usage or script will complain. With this forward declaration we can use them anywhere and main code can be at the script beginning. I would prefer to keep it this way, since its like this in all our scripts.

This was maybe true years ago, but now it works without the forward declaration (maybe unless it's a prototype). I think this way you are defining two subs but the second one overwrites the first one. However, I know we have that in all our scripts so let's keep it that way for now.

balcirakpeter (approved):

PR #497, opened 2020-11-02
Migrate services to getHashedData methods

stavamichal:
- migrate ad_mu and o365_mu to use GenHasedData class (in perl) instead
 of ServiceAttributes. Calling new method getHashedDataWithGroups.
 - for ad_mu: use virtual adDisplayName instead of defined one (for old
 behavior it should work the same way)

### DIFF
--- a/gen/ad_mu
+++ b/gen/ad_mu
@@ -27,7 +27,7 @@ my $fileNameLicGroupNames = "$DIRECTORY/licGroupNames";
 my $fileNameFID = "$DIRECTORY/facilityId";
 my $fileNameManualEmployees = "$DIRECTORY/manualEmployees";
 
-my $data = perunServicesInit::getDataWithGroups;
+my $data = perunServicesInit::getHashedDataWithGroups;
 
 # Constants
 our $A_F_BASE_DN;  *A_F_BASE_DN = \'urn:perun:facility:attribute-def:def:adBaseDN';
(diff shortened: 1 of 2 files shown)


balcirakpeter (approved):
We have to decide what to do with the adDisplayName attribute. Otherwise, it seems to be ok.

stavamichal:
If displayName is empty, we will use adName instead.

stavamichal:
It should be fixed now.

balcirakpeter (approved):

gen/ad_mu
```diff
@@ -518,19 +516,19 @@ for my $login (@logins) {
...
 
-	if($users->{$login}->{$A_U_F_O365_PREFERRED_LANGUAGE}) {
+	if ($users->{$login}->{$A_U_F_O365_PREFERRED_LANGUAGE}) {
 		print FILE "preferredLanguage: " . $users->{$login}->{$A_U_F_O365_PREFERRED_LANGUAGE} . "\n";
 	}
 
 	print FILE "targetaddress: " . $login . '@mail.muni.cz' . "\n";
 
-	print FILE "msDS-cloudExtensionAttribute1: " . $login . '@' . $facilityAttributes{$A_F_AZURE_DOMAIN} . "\n";
+	print FILE "msDS-cloudExtensionAttribute1: " . $login . '@' . $domain . "\n";
```
zlamalp, line 525: ``$domain`` contains ``A_F_DOMAIN`` and not ``A_F_AZURE_DOMAIN`` after this change. Is it ok?

stavamichal: Nice catch. You are right. Fixed.

stavamichal: We didn't notice because there are the same values in both attributes at this moment in testing data.

zlamalp (approved):

PR #499, opened 2020-11-03
do not specify private ssh key directly in generic_send

mvocu:
Do not use hardwired name of perun's private ssh key in generic_send, as the key name may vary between instances/os versions. 

It is still possible to override the key name by specifying TRANSPORT_COMMAND in service configuration.

### DIFF
--- a/send/generic_send
+++ b/send/generic_send
@@ -11,7 +11,6 @@ TIMEOUT_KILL="60" # 60 sec to kill after timeout
 FACILITY_NAME=$1
 DESTINATION=$2
 DESTINATION_TYPE=$3
-KEY_PATH="`echo ~`/.ssh/id_rsa"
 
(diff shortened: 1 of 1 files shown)


balcirakpeter (approved):

zlamalp (commented):
I would prefer having ``KEY_PATH`` configurable and append it to the default ``TRANSPORT_COMMAND`` if not empty. It will be easier to just specify the key without the whole command if the default command changes for some reason.
But I know only about two services at VŠUP which do that and they both override both KEY_PATH and TRANSPORT_COMMAND, so it will work.

mvocu:
I'd rather not do that. It would only complicate the interface by adding second way of doing the same thing.

zlamalp (approved):

PR #505, opened 2020-12-09
Use getHashedHierarchicalData() in appdb service

zlamalp:
- Use new getHashedHierarchicalData() in appdb service, since
  it now contains VO related attributes.

### DIFF
--- a/gen/appdb
+++ b/gen/appdb
@@ -9,11 +9,11 @@ use XML::Simple;
 
 local $::SERVICE_NAME = "appDB";
 local $::PROTOCOL_VERSION = "3.0.0";
-my $SCRIPT_VERSION = "3.0.0";
+my $SCRIPT_VERSION = "3.1.0";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
-my $data = perunServicesInit::getDataWithVos;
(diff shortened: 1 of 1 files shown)


gen/appdb
```diff
@@ -38,88 +38,104 @@ my $vosStruc = {};
...
-									"dn" => \@dn,
-			          };
+		my $appDbContactRole = $data->getResourceAttributeValue(resource => $resourceId, attrName => $A_R_APPDB_CONTACT_ROLE);
 
+		my $voId = $data->getVoIdForResource(resource => $resourceId);
+		my $voShortName = $data->getVoAttributeValue(vo => $voId, attrName => $A_VO_SHORT_NAME);
 
-			#add user to global structure
-			push @{$usersXml->{"result"}}, { "row" => \@row };
+		my $userId = $data->getUserAttributeValue(member => $memberId, attrName => $A_USER_ID);
```
stavamichal, line 103: This could be done by:
```
$data->getUserIdForMember( member => $memberId );
```

stavamichal (approved):

gen/appdb
zlamalp: You are right, I fixed it. We can generally remove all ``entity:core:id`` attributes from required once each service is converted to new hashed data structures.

stavamichal (approved):

PR #506, opened 2020-12-09
Sort mails within the list in mailman like services

zlamalp:
- We want to compare script results when script implementation
  changes, so we should sort output (list of mails within
  each mailing list).


stavamichal (approved):

zlamalp:
Changes were included in gen scripts on rewrites to usage of hashed data. Closing this pull-request.

PR #509, opened 2021-01-12
vsup_dc2 edited to use getHashedData

balcirakpeter:
- For this purpose was used method getHashedHierarchicalData, which
  aggregates all members under the facility.


zlamalp (approved):
Can you please update also ``$SCRIPT_VERSION`` to eg. ``3.0.1``. Otherwise it works, thank you.

PR #511, opened 2021-01-12
vsup_k4 edited to use getHashedData

balcirakpeter:
- For this purpose was used method getHashedDataWithGroups.


zlamalp (approved):
Please just update script version. Thank you.

PR #518, opened 2021-01-13
vsup_zimbra edited to use getHashedData

balcirakpeter:

### DIFF
--- a/gen/vsup_zimbra
+++ b/gen/vsup_zimbra
@@ -12,17 +12,16 @@ sub getStatus;
 
 local $::SERVICE_NAME = "vsup_zimbra";
 local $::PROTOCOL_VERSION = "3.1.0";
-my $SCRIPT_VERSION = "3.0.1";
+my $SCRIPT_VERSION = "3.0.2";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
 my $fileName = "$DIRECTORY/$::SERVICE_NAME".".csv";
-my $data = perunServicesInit::getHierarchicalData;
+my $data = perunServicesInit::getHashedHierarchicalData;
 my $agent = Perun::Agent->new();
 my $attributesAgent = $agent->getAttributesAgent;
 
 # Constants
-our $A_USER_ID; *A_USER_ID = \'urn:perun:user:attribute-def:core:id';
 our $A_LOGIN; *A_LOGIN = \'urn:perun:user_facility:attribute-def:virt:login';
 our $A_UCO; *A_UCO = \'urn:perun:user:attribute-def:def:ucoVsup';
 our $A_FIRST_NAME;  *A_FIRST_NAME = \'urn:perun:user:attribute-def:core:firstName';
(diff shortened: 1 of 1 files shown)


gen/vsup_zimbra
```diff
@@ -54,32 +53,44 @@ my $users;  # $users->{$uco}->{ATTR} = $attrValue;
...
+		my $expirationManual = $data->getUserAttributeValue(member => $memberId, attrName => $A_EXPIRATION_MANUAL);
+		my $vsupMailAlias = $data->getUserAttributeValue(member => $memberId, attrName => $A_EMAIL_VSUP_ALIAS);
+		my $vsupMailAliases = $data->getUserAttributeValue(member => $memberId, attrName => $A_EMAIL_VSUP_ALIASES);
+
 
-		$users->{$uco}->{$A_USER_ID} = $uAttributes{$A_USER_ID};
+		$users->{$uco}->{$A_LOGIN} = $login;
+		$users->{$uco}->{'EMAIL'} = ($vsupMail || $login . '@vsup.cz');
+
+		$users->{$uco}->{$A_USER_ID} = $userId;
```
zlamalp, line 93: Can be replaced with ``getUserIdForMember()`` and required attribute ``user:core:id`` can be removed.

gen/vsup_zimbra
```diff
@@ -54,32 +53,44 @@ my $users;  # $users->{$uco}->{ATTR} = $attrValue;
...
+		my $uFirstName = $data->getUserAttributeValue(member => $memberId, attrName => $A_FIRST_NAME);
+		my $artisticLastName = $data->getUserAttributeValue(member => $memberId, attrName => $A_ARTISTIC_LAST_NAME);
+		my $uLastName = $data->getUserAttributeValue(member => $memberId, attrName => $A_LAST_NAME);
+		my $uDisplayName = $data->getUserAttributeValue(member => $memberId, attrName => $A_ZIMBRA_DISPLAY_NAME);
+		my $blacklisted = $data->getUserFacilityAttributeValue(member => $memberId, attrName => $A_BLACKLISTED);
+		my $expirationKOS = $data->getUserAttributeValue(member => $memberId, attrName => $A_EXPIRATION_KOS);
+		my $expirationDC2 = $data->getUserAttributeValue(member => $memberId, attrName => $A_EXPIRATION_DC2);
+		my $expirationManual = $data->getUserAttributeValue(member => $memberId, attrName => $A_EXPIRATION_MANUAL);
+		my $vsupMailAlias = $data->getUserAttributeValue(member => $memberId, attrName => $A_EMAIL_VSUP_ALIAS);
+		my $vsupMailAliases = $data->getUserAttributeValue(member => $memberId, attrName => $A_EMAIL_VSUP_ALIASES);
```
zlamalp, line 87: Following attributes are manually retrieved from RPC and they are not present in hashed data structure.

gen/vsup_zimbra
```diff
@@ -142,43 +153,19 @@ foreach my $rData ($data->getChildElements) {
...
-			if ($aliases) {
-				@aliases = @$aliases;
-			}
-			$users->{$uco}->{$A_EMAIL_VSUP_ALIASES} = join(",",@aliases) || '';
-
-		}
-	}
-
-}
-
```
zlamalp, line 181: We must keep manual retrieval of those 2 attributes from the RPC, since we don't want the service to require those attributes (generate them on perun side), but we must consistently propagate them with the rest of params to the Zimbra using this service.

gen/vsup_zimbra
```diff
@@ -110,23 +121,23 @@ foreach my $rData ($data->getChildElements) {
...
-				die "User: '$uAttributes{$A_LOGIN}' was both MAILMAN-LIST type and $users->{$uco}->{'TYPE'} type.";
+				die "User: '$login' was both MAILMAN-LIST type and $users->{$uco}->{'TYPE'} type.";
 			}
 		}
 
 		# get users name
-		my $firstName = ($uAttributes{$A_ARTISTIC_FIRST_NAME} || ($uAttributes{$A_FIRST_NAME} || ''));
-		my $lastName = ($uAttributes{$A_ARTISTIC_LAST_NAME} || ($uAttributes{$A_LAST_NAME} || ''));
+		my $firstName = ($artisticFirstName || ($uFirstName || ''));
+		my $lastName = ($artisticLastName || ($uLastName || ''));
```
zlamalp, line 138: It would be nice, if this could be simplified, so we wouldn't need new ``$uFirstName`` or ``$uLastName`` variables. Maybe like this (for firstName):
```perl
my $firstName = $data->getUserAttributeValue(member => $memberId, attrName => $A_FIRST_NAME);
$firstName = ($artisticFirstName || ($firstName || ''));
```

I'm not sure about later ``$uDisplayName`` with ``$uDisplayName || undef;``. I think its meant to handle cases, when display name is defined but empty from Perun for following condition. If we update later ``unless`` condition, we can also use simple ``$displayName``, like:

```perl
my $displayName = $data->getUserAttributeValue(member => $memberId, attrName => $A_ZIMBRA_DISPLAY_NAME);
$displayName = $displayName || '';
unless(defined $displayName and length $displayName) {
    # construct display name from partial params
}
```
Then we could be sure, that at the end, all variables are defined:
```perl
$users->{$uco}->{"FIRST_NAME"} = $firstName;
$users->{$uco}->{"LAST_NAME"} = $lastName;
$users->{$uco}->{"DISPLAY_NAME"} = $displayName;
```

zlamalp (changes requested):
Please update also script version.

gen/vsup_zimbra
balcirakpeter: Fixed

gen/vsup_zimbra
balcirakpeter: Fixed

gen/vsup_zimbra
balcirakpeter: Fixed

gen/vsup_zimbra
balcirakpeter: Fixed

zlamalp (approved):

PR #521, opened 2021-01-14
mailman edited to use getHashedData

balcirakpeter:

### DIFF
--- a/gen/mailman
+++ b/gen/mailman
@@ -11,11 +11,11 @@ use POSIX qw/strftime/;
 
 local $::SERVICE_NAME = "mailman";
 local $::PROTOCOL_VERSION = "3.1.0";
-my $SCRIPT_VERSION = "3.2.0";
+my $SCRIPT_VERSION = "3.2.1";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
-my $data = perunServicesInit::getHierarchicalData;
+my $data = perunServicesInit::getHashedHierarchicalData;
 
 #Constants
 our $A_USER_NAME;                        *A_USER_NAME =                       \'urn:perun:user:attribute-def:core:displayName';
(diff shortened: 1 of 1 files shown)


gen/mailman
```diff
@@ -35,26 +35,25 @@ my $mailinglistsDirectory = "$DIRECTORY/mailinglists";
...
-    foreach my $memberAttributes (dataToAttributesHashes @membersData) {
-        #list only VALID members except if allowNonvalidUsers is true
-        next unless($memberAttributes->{$A_USER_STATUS} eq "VALID");
+		my $userStatus = $data->getMemberAttributeValue(member => $memberId, attrName => $A_USER_STATUS);
+		my $userName = $data->getUserAttributeValue(member => $memberId, attrName => $A_USER_NAME);
+		my $userMail = $data->getUserAttributeValue(member => $memberId, attrName => $A_USER_MAIL);
 
-        $mailinglistStruc->{$listName}->{$memberAttributes->{$A_USER_MAIL}}->{$A_USER_NAME} = $memberAttributes->{$A_USER_NAME};
-    }
+		#list only VALID members except if allowNonvalidUsers is true
```
zlamalp, line 52: This comment is not true, since there is no checking of ``allowNonvalidUsers`` attribute.

gen/mailman
```diff
@@ -65,7 +64,7 @@ for my $listName (keys %$mailinglistStruc) {
     # As a first line, print header
```
zlamalp, line 64: I'm just wondering if we should sort also the lists themselves, so that if generating one of them fails, we can reproduce also the order in which they are generated. But its probably not needed.

zlamalp (changes requested):

gen/mailman
balcirakpeter: Fixed

zlamalp (approved):

PR #522, opened 2021-01-14
mailman_elixir edited to use getHashedData

balcirakpeter:

### DIFF
--- a/gen/mailman_elixir
+++ b/gen/mailman_elixir
@@ -11,11 +11,11 @@ use POSIX qw/strftime/;
 
 local $::SERVICE_NAME = "mailman_elixir";
 local $::PROTOCOL_VERSION = "3.1.0";
-my $SCRIPT_VERSION = "3.2.0";
+my $SCRIPT_VERSION = "3.2.1";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
-my $data = perunServicesInit::getHierarchicalData;
+my $data = perunServicesInit::getHashedHierarchicalData;
 
 #Constants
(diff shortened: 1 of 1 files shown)


gen/mailman_elixir
```diff
@@ -36,27 +36,28 @@ mkdir $mailinglistsDirectory or die "Can't mkdir $mailinglistsDirectory: $!";
...
-    foreach my $memberAttributes (dataToAttributesHashes @membersData) {
-        #list only VALID members except if allowNonvalidUsers is true
-        next unless($memberAttributes->{$A_USER_STATUS} eq "VALID");
+		my $memberStatus = $data->getMemberAttributeValue(member => $memberId, attrName => $A_USER_STATUS);
+		my $userMail = $data->getUserAttributeValue(member => $memberId, attrName => $A_USER_MAIL);
+		my $userName = $data->getUserAttributeValue(member => $memberId, attrName => $A_USER_NAME);
 
-        $mailinglistStruc->{$listName}->{$memberAttributes->{$A_USER_MAIL}}->{$A_USER_NAME} = $memberAttributes->{$A_USER_NAME};
-    }
+		#list only VALID members except if allowNonvalidUsers is true
```
zlamalp, line 56: I don't think, that we process ``allowNonvalidUsers`` anywhere in the script, so the comment is misleading.

zlamalp (changes requested):

gen/mailman_elixir
balcirakpeter: Fixed

zlamalp (approved):

PR #540, opened 2021-01-20
Don't show the progress of curl cmd

vyskocilpavel:
* Don't show the progress of the curl commands in 'bbmri_collections' and 'bbmri_networks' send script => use curl with '-s'

### DIFF
--- a/send/bbmri_collections
+++ b/send/bbmri_collections
@@ -26,7 +26,7 @@ fi
 [ -r /etc/perun/services/$SERVICE_NAME ] && . /etc/perun/services/$SERVICE_NAME
 
 # Specify command to transfer the data
-TRANSPORT_COMMAND="curl -i -H Content-Type:application/json -f -X POST"
+TRANSPORT_COMMAND="curl -Ss -i -H Content-Type:application/json -f -X POST"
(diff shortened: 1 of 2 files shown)


send/bbmri_collections
```diff
@@ -26,7 +26,7 @@ fi
 [ -r /etc/perun/services/$SERVICE_NAME ] && . /etc/perun/services/$SERVICE_NAME
 
 # Specify command to transfer the data
-TRANSPORT_COMMAND="curl -i -H Content-Type:application/json -f -X POST"
+TRANSPORT_COMMAND="curl -s -i -H Content-Type:application/json -f -X POST"
```
zlamalp, line 29: I would prefer using ``-Ss`` so that curl is silent, until there is an error. Right know it would silent also the error output, not just the "progress". For Debian 11 (which is not case of BBMRI instance) we could also use only ``--no-progress-meter`` which hides just the progress without silencing other warnings or errors.

zlamalp (changes requested):

send/bbmri_collections
vyskocilpavel: Done

zlamalp (approved):

PR #542, opened 2021-01-20
Service fs_replicas is now using getHashedData

stavamichal:
- there is no change in the generated data, just the way of generating
is now different, simplier and quicker


cuadradek (changes requested):
The committed file is fs_quotas, not fs_replicas.

PR #545, opened 2021-01-20
Service fs_replicas is now using getHashedData

stavamichal:
- there is no change in the generated data, just the way of generating
  is now different, simplier and quicker

### DIFF
--- a/gen/fs_replicas
+++ b/gen/fs_replicas
@@ -11,15 +11,14 @@ sub processMemberData;
 
 local $::SERVICE_NAME = "fs_replicas";
 local $::PROTOCOL_VERSION = "3.0.0";
-my $SCRIPT_VERSION = "3.0.0";
+my $SCRIPT_VERSION = "3.0.1";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
-my $data = perunServicesInit::getDataWithGroups;
+my $data = perunServicesInit::getHashedDataWithGroups;
 
 #Constants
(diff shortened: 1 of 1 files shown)


gen/fs_replicas
```diff
@@ -11,15 +11,15 @@ sub processMemberData;
...
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
-my $data = perunServicesInit::getDataWithGroups;
+my $data = perunServicesInit::getHashedDataWithGroups;
 
 #Constants
-our $A_USER_LOGIN;                              *A_USER_LOGIN =                              \'urn:perun:user_facility:attribute-def:virt:login';
-our $A_USER_LOGIN_NAMESPACE;                    *A_USER_LOGIN_NAMESPACE =                    \'urn:perun:facility:attribute-def:def:loginNamespace';
+our $A_UF_LOGIN;                                *A_UF_LOGIN =                                \'urn:perun:user_facility:attribute-def:virt:login';
+our $A_UF_LOGIN_NAMESPACE;                      *A_UF_LOGIN_NAMESPACE =                      \'urn:perun:facility:attribute-def:def:loginNamespace';
```
cuadradek, line 22: $A_UF_LOGIN_NAMESPACE should be renamed to $A_U_LOGIN_NAMESPACE because it is facility attribute.

cuadradek (changes requested):

gen/fs_replicas
stavamichal: You are right, I removed it because it is an unused variable at this moment.

cuadradek (approved):

PR #549, opened 2021-01-28
Remove sorting of emails for resource-mails in o365_mu

stavamichal:
- there is order of emails we need to use from the Perun (for this
 reason we can't sort emails here)


rosinec:
Script is now generating valid order of email adresses for mailboxes in O365.

PR #550, opened 2021-02-03
All sshkeys services edited to use getHashedData

HejdaJakub:
* Services sshkeys, sshkeys_generic and sshkeys_root are now using getHashedHierarchicalData.

### DIFF
--- a/gen/sshkeys
+++ b/gen/sshkeys
@@ -7,35 +7,25 @@ use File::Basename;
 
 our $SERVICE_NAME = basename($0);
 our $PROTOCOL_VERSION = "3.1.0";
-my $SCRIPT_VERSION = "3.0.1";
+my $SCRIPT_VERSION = "3.0.2";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
-my $data = perunServicesInit::getHierarchicalData;
+my $data = perunServicesInit::getHashedHierarchicalData;
 
 my $sshkeysDirectory = "$DIRECTORY/sshkeys";
 mkdir $sshkeysDirectory or die "Can't mkdir $sshkeysDirectory: $!";
 
 #Constants
 our $A_SSHKEYS;                      *A_SSHKEYS =                     \'urn:perun:user:attribute-def:def:sshPublicKey';
-our $A_USER_LOGIN;                   *A_USER_LOGIN =                  \'urn:perun:user_facility:attribute-def:virt:login';
-
-my $service_file_name = "$DIRECTORY/$::SERVICE_NAME";
+our $A_USER_FACILITY_LOGIN;          *A_USER_FACILITY_LOGIN =         \'urn:perun:user_facility:attribute-def:virt:login';
 
 my %sshKeys;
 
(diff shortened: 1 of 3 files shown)


gen/sshkeys
```diff
@@ -7,34 +7,26 @@ use File::Basename;
...
-
-my @resourcesData = $data->getChildElements;
-foreach my $rData (@resourcesData) {
-	my %values;
-	my @membersData = $rData->getChildElements;
-	foreach my $mData (@membersData) {
-		my %memberAttributes = attributesToHash $mData->getAttributes;
-
-		$sshKeys{$memberAttributes{$A_USER_LOGIN}} = $memberAttributes{$A_SSHKEYS};
+foreach my $resourceId ($data->getResourceIds()) {
```
balcirakpeter, line 25: These two for each cycles can be merged into one by using the method `getMemberIdsForFacility()`. Sorry that I did not mention it earlier. I forgot that we have this method.

gen/sshkeys_root
```diff
@@ -20,23 +20,17 @@ my $service_file_name = "$DIRECTORY/$::SERVICE_NAME";
 
 my @sshKeys;
 
-my %facilityAttributes = attributesToHash $data->getAttributes;
-#####################################
+foreach my $resourceId ($data->getResourceIds()) {
```
balcirakpeter, line 23: Same as above.

balcirakpeter (changes requested):

gen/sshkeys_root
HejdaJakub: Both scripts have been edited.

balcirakpeter (approved):

gen/sshkeys_generic
```diff
@@ -8,44 +8,42 @@ use File::Basename;
...
-foreach my $resourceData (@resourcesData) {
-	my %resourceAttributes = attributesToHash $resourceData->getAttributes;
-	#get information about name of target user (file name for data)
-	my $fileName = $resourceAttributes{$A_RESOURCE_SSHKEYS_TARGET_USER};
-	foreach my $memberData ($resourceData->getChildElements) {
-		my %memberAttributes = attributesToHash $memberData->getAttributes;
+foreach my $resourceId ($data->getResourceIds()) {
+	my $fileName = $data->getResourceAttributeValue(resource => $resourceId, attrName => $A_RESOURCE_SSHKEYS_TARGET_USER);
+	foreach my $memberId ($data->getMemberIdsForResource( resource => $resourceId )) {
+		my $userStatus = $data->getMemberAttributeValue(member => $memberId, attrName => $A_MEMBER_STATUS);
```
zlamalp, line 29: Since you changed ``A_USER_STATUS`` to ``A_MEMBER_STATUS`` I would suggest to change also destination variable from ``userStatus`` to ``memberStatus``.

zlamalp (changes requested):
Just one small think. Otherwise its ok. Thank you.

gen/sshkeys_generic
HejdaJakub: Good point, I didn't notice that variable. Changed.

zlamalp (approved):

PR #554, opened 2021-02-05
voms, voms_dirac, du_users_export edited to use hashed data

cuadradek:
-Services voms, voms_dirac and du_users_export are now using getHashedData.


mvocu (approved):
Good catch with the attribute names.

stavamichal (approved):

PR #556, opened 2021-02-05
 ad_user_mu_ucn service refactored and added new features

rosinec:
- migrated to getHashedData
 - we are sending chip number to AD
 - added checkBase64 function from safeq script
 - grace period mechanism (allow_not_valid_members) - allow users that are not valid for time defined by group membership expiration
 - take users from each resource - only users assigned to resource are propagated to AD
 - displayName changes - academic prefix or suffix is not needed
 - allowing all users logins - not only "9UCO"

### DIFF
--- a/gen/ad_user_mu_ucn
+++ b/gen/ad_user_mu_ucn
@@ -34,9 +34,6 @@ our $A_R_NAME; *A_R_NAME = \'urn:perun:resource:attribute-def:core:name';
 our $STATUS_VALID; *STATUS_VALID = \'VALID';
 our $STATUS_EXPIRED; *STATUS_EXPIRED = \'EXPIRED';
 
-# Get the facility ID
-my $facilityId = $data->getFacilityId();
-
 # CHECK ON FACILITY ATTRIBUTES
 if (!defined($data->getFacilityAttributeValue( attrName => $A_F_BASE_DN ))) {
 	exit 1;
@@ -75,7 +72,7 @@ foreach my $resourceId ( $data->getResourceIds() ){
 	foreach my $memberId ($data->getMemberIdsForResource( resource => $resourceId )) {
 
 		my $login = $data->getUserFacilityAttributeValue( member => $memberId, attrName => $A_LOGIN );
-			my $memberStatus = $data->getMemberAttributeValue( member => $memberId, attrName => $A_MEMBER_STATUS );
+		my $memberStatus = $data->getMemberAttributeValue( member => $memberId, attrName => $A_MEMBER_STATUS );
 		#next unless ( $uAttributes{$A_MEMBER_STATUS} eq $STATUS_VALID || ( $uAttributes{$A_MEMBER_STATUS} eq $STATUS_EXPIRED && $rAttributes{$ALLOW_NOT_VALID_MEMBERS}));
 		next unless ( ($memberStatus eq $STATUS_VALID) || ( ($memberStatus eq $STATUS_EXPIRED) && $allowNotValidMembers));
 
--- a/gen/ad_user_mu_ucn
+++ b/gen/ad_user_mu_ucn
@@ -4,6 +4,9 @@ use strict;
 use warnings;
 use perunServicesInit;
 use perunServicesUtils;
+use MIME::Base64;
+use utf8;
+use Encode;
 
 local $::SERVICE_NAME = "ad_user_mu_ucn";
 local $::PROTOCOL_VERSION = "3.0.0";
@@ -14,7 +17,11 @@ my $DIRECTORY = perunServicesInit::getDirectory;
 my $fileName = "$DIRECTORY/$::SERVICE_NAME".".ldif";
 my $baseDnFileName = "$DIRECTORY/baseDN";
 
+<<<<<<< HEAD
 my $data = perunServicesInit::getHashedHierarchicalData;
+=======
+my $data = perunServicesInit::getHashedHierarchicalData(1);
+>>>>>>> ad_user_mu_ucn service refactored and added new features
 
 #Constants
 our $A_F_BASE_DN;  *A_F_BASE_DN = \'urn:perun:facility:attribute-def:def:adBaseDN';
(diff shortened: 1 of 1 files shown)


rosinec:
Also matched requirements for migrating from AccMgr at MU.

gen/ad_user_mu_ucn
```diff
@@ -113,6 +133,10 @@ for my $login (@logins) {
 	if (defined $mail and length $mail) {
 		print FILE "mail: " . $mail . "\n";
 	}
+	foreach (@{$users->{$login}->{$A_CHIPNUMBERS}}){
+		next if ($_ eq "");
+		print FILE "houseIdentifier: " . checkBase64($_) . "\n";
```
zlamalp, line 138: You shouldn't have ending with space, since you optionally append ``:`` and base64 value, so you end up with ``houseIdentifier: :somevalue``, which I think is not valid. Spaces between separator ``: or ::`` and value are optional, so you can discard it in this not and don't have to implement any additional logic.

gen/ad_user_mu_ucn
```diff
@@ -127,4 +151,13 @@ for my $login (@logins) {
 
 close(FILE);
 
+sub checkBase64 {
```
zlamalp, line 154: I believe that all subs needs to be defined at the top like: ``sub checkBase64;`` aka - before their actual usage, but it doesn't seem to be a case even in the ``safeq`` service, from which I believe this code originates. So if it works for you, then its OK.

gen/ad_user_mu_ucn
```diff
@@ -52,23 +65,35 @@ open FILE,">:encoding(UTF-8)","$baseDnFileName" or die "Cannot open $baseDnFileN
...
+
+		unless ($login =~ /^s-/){
+			$users->{$login}->{"DN"} = "CN=" . $login . "," . $baseDN;
+		} else {
+			$users->{$login}->{"DN"} = "CN=" . $login . "," . $serviceAccountsDN;
+		}
+		# store standard attrs
+		$users->{$login}->{$A_FIRST_NAME} = $data->getUserAttributeValue(member => $memberId, attrName => $A_FIRST_NAME);
+		$users->{$login}->{$A_LAST_NAME} = $data->getUserAttributeValue(member => $memberId, attrName => $A_LAST_NAME);
+		$users->{$login}->{$A_DISPLAY_NAME} = $data->getUserAttributeValue(member => $memberId, attrName => $A_DISPLAY_NAME);
```
zlamalp, line 93: It seems, that after you changes, ``user:core:displayName`` attribute is no longer used and can be removed from the script and from set of required attributes once deployed.

gen/ad_user_mu_ucn
```diff
@@ -98,7 +118,7 @@ for my $login (@logins) {
 	# skip attributes which are empty and LDAP can't handle it (FIRST_NAME, EMAIL)
 	my $sn = $users->{$login}->{$A_LAST_NAME};
 	my $givenName = $users->{$login}->{$A_FIRST_NAME};
-	my $displayName = $users->{$login}->{$A_DISPLAY_NAME};
+	my $displayName = ($users->{$login}->{$A_FIRST_NAME} || "") . " " . ($users->{$login}->{$A_LAST_NAME} || "");
```
zlamalp, line 121: Since firstName and lastName are generally not required to be non-empty, is it OK to have empty space as a displayName value?

gen/ad_user_mu_ucn
```diff
@@ -52,23 +65,35 @@ open FILE,">:encoding(UTF-8)","$baseDnFileName" or die "Cannot open $baseDnFileN
...
-	# store standard attrs
-	$users->{$login}->{$A_FIRST_NAME} = $uAttributes{$A_FIRST_NAME};
-	$users->{$login}->{$A_LAST_NAME} = $uAttributes{$A_LAST_NAME};
-	$users->{$login}->{$A_DISPLAY_NAME} = $uAttributes{$A_DISPLAY_NAME};
-	$users->{$login}->{$A_MAIL} = $uAttributes{$A_MAIL};
-
+# FOR EACH RESOURCE
+foreach my $resourceId ( $data->getResourceIds() ){
+        
+        my $allowNotValidMembers = $data->getResourceAttributeValue( resource => $resourceId, attrName => $ALLOW_NOT_VALID_MEMBERS);
```
zlamalp, line 76: Indentation, please use tabs.

gen/ad_user_mu_ucn
```diff
@@ -52,23 +65,35 @@ open FILE,">:encoding(UTF-8)","$baseDnFileName" or die "Cannot open $baseDnFileN
 print FILE $baseDN;
 close(FILE);
 
+#my %allChipNumbers;  #used for consistency check - each chip number should belong to only one user
```
zlamalp, line 68: I'm not sure, what are the constraints in AD, but having this check implemented would be generally better, so either implement it or remove the unused comment.

zlamalp (changes requested):

stavamichal:
@rosinec Please add all missing information about changes to the commit message.

gen/ad_user_mu_ucn
```diff
@@ -14,33 +15,42 @@ my $DIRECTORY = perunServicesInit::getDirectory;
...
+our $A_MEMBER_STATUS; *A_MEMBER_STATUS = \'urn:perun:member:attribute-def:core:status';
+our $ALLOW_NOT_VALID_MEMBERS; *ALLOW_NOT_VALID_MEMBERS = \'urn:perun:resource:attribute-def:def:AllowNotValidMembers';
+
+our $A_R_NAME; *A_R_NAME = \'urn:perun:resource:attribute-def:core:name';
+
+our $STATUS_VALID; *STATUS_VALID = \'VALID';
+our $STATUS_EXPIRED; *STATUS_EXPIRED = \'EXPIRED';
+
+# Get the facility ID
+my $facilityId = $data->getFacilityId();
```
zlamalp, line 38: Unused variable, please remove it.

gen/ad_user_mu_ucn
```diff
@@ -113,6 +140,10 @@ for my $login (@logins) {
 	if (defined $mail and length $mail) {
 		print FILE "mail: " . $mail . "\n";
 	}
+	foreach (@{$users->{$login}->{$A_CHIPNUMBERS}}){
```
zlamalp, line 143: I'm not sure, but I think this foreach with re-typing to array will fail if chip numbers attribute is `undef`. It can be, since I don't see any check (attribute module) in perun core. I think you should add check above like:
```perl
my $chipNumbers = $users->{$login}->{$A_CHIPNUMBERS};
if (defined $chipNumbers) {
    # ... LDIF print code ...
}
```

zlamalp (changes requested):
Please rebase the pull-request. We migrated the original script to usage of hashed data and now it conflicts, but your version can be safely used as it is, since you migrated too.

gen/ad_user_mu_ucn
zlamalp: This is still true, `$facilityId` variable is not used within the script and can be removed.

gen/ad_user_mu_ucn
```diff
@@ -51,21 +62,34 @@ open FILE,">:encoding(UTF-8)","$baseDnFileName" or die "Cannot open $baseDnFileN
...
-
+# FOR EACH RESOURCE
+foreach my $resourceId ( $data->getResourceIds() ){
+
+	my $allowNotValidMembers = $data->getResourceAttributeValue( resource => $resourceId, attrName => $ALLOW_NOT_VALID_MEMBERS);
+	# EACH USER ON RESOURCE
+	foreach my $memberId ($data->getMemberIdsForResource( resource => $resourceId )) {
+
+		my $login = $data->getUserFacilityAttributeValue( member => $memberId, attrName => $A_LOGIN );
+			my $memberStatus = $data->getMemberAttributeValue( member => $memberId, attrName => $A_MEMBER_STATUS );
```
zlamalp, line 78: indentation

zlamalp (approved):
Seems to be ok, just needs a little cleanup.

PR #560, opened 2021-02-11
Fixed sorting of output in pbs_publication_fairshare

zlamalp:
- Sort output also by name/login within each weight
  group. Since weight is rounded when printed to file,
  entries doesn't seems to be sorted, but they are.


zlamalp:
I'm merging this, since this version is already deployed on single instance it is actually used.

PR #561, opened 2021-02-12
Services ceitec*, crm* and crocs* edited to use getHashedData

HejdaJakub:
* Services ceitec_proteomics and crm_ceitec are now using getHashedDataWithGroups.
* Service crocs_user_export is now using getHashedHierarchicalData.

### DIFF
--- a/gen/crocs_user_export
+++ b/gen/crocs_user_export
@@ -8,42 +8,41 @@ use JSON::XS;
 
 our $SERVICE_NAME = "crocs_user_export";
 our $PROTOCOL_VERSION = "3.0.0";
-my $SCRIPT_VERSION = "3.0.0";
+my $SCRIPT_VERSION = "3.0.1";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
-my $data = perunServicesInit::getHierarchicalData;
+my $data = perunServicesInit::getHashedHierarchicalData;
 
 #Constants
 our $A_USER_FIRSTNAME;            *A_USER_FIRSTNAME =        \'urn:perun:user:attribute-def:core:firstName';
 our $A_USER_LASTNAME;             *A_USER_LASTNAME =         \'urn:perun:user:attribute-def:core:lastName';
 our $A_USER_LOGIN;                *A_USER_LOGIN=             \'urn:perun:user:attribute-def:def:login-namespace:mu';
 
-my %attributesByLogin;
+my $attributesByLogin;
 
-my @resourcesData = $data->getChildElements;
-foreach my $rData (@resourcesData) {
-	my @membersData = $rData->getChildElements;
-	foreach my $mData (@membersData) {
-		my %memberAttributes = attributesToHash $mData->getAttributes;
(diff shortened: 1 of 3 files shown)


gen/crocs_user_export
```diff
@@ -8,36 +8,35 @@ use JSON::XS;
...
 my @users;
-for my $login (sort keys %attributesByLogin) {
-	my $values = $attributesByLogin{$login};
+for my $login (sort keys %{$attributesByLogin}) {
+	my $values = $attributesByLogin->{$login};
 	my $user = {};
 	$user->{"firstname"} = $values->{$A_USER_FIRSTNAME} || "";
 	$user->{"lastname"}  = $values->{$A_USER_LASTNAME} || "";
-	$user->{"UCO"}       = $values->{$A_USER_LOGIN};
+	$user->{"UCO"}       = $values;
```
zlamalp, line 38: You should print login (aka UCO) here and not the whole `$values` structure. Previously it printed object like:
```json
{
      "firstname" : "GivenName",
      "UCO" : "123456",
      "lastname" : "Surname"
}
```
Now it prints:
```json
{
      "lastname" : "Surname",
      "UCO" : {
         "urn:perun:user:attribute-def:core:firstName" : "GivenName",
         "urn:perun:user:attribute-def:core:lastName" : "Surname"
      },
      "firstname" : "GivenName"
}
```

zlamalp (changes requested):

gen/crocs_user_export
zlamalp: Also please use `canonical` while printing JSON, so that objects have sorted keys and results are comparable between the runs, like: 
```perl
JSON::XS->new->utf8->pretty->canonical->encode(\@users);
```

gen/crocs_user_export
```diff
@@ -8,42 +8,42 @@ use JSON::XS;
...
 my @users;
-for my $login (sort keys %attributesByLogin) {
-	my $values = $attributesByLogin{$login};
+for my $login (sort keys %{$attributesByLogin}) {
+	my $values = $attributesByLogin->{$login};
 	my $user = {};
 	$user->{"firstname"} = $values->{$A_USER_FIRSTNAME} || "";
 	$user->{"lastname"}  = $values->{$A_USER_LASTNAME} || "";
-	$user->{"UCO"}       = $values->{$A_USER_LOGIN};
+	$user->{"UCO"}       = $values>{$A_USER_LOGIN};
```
zlamalp, line 39: Missing `-` in `$values>{$A_USER_LOGIN};`, but when I think about it, we can simplify it further. We can use simple `$user->{"UCO"} = $login`.
Because we already have value for login as a key in the main hash. There is no need to store it also under the `$A_USER_LOGIN` key within the stored user object.

zlamalp (changes requested):

gen/crocs_user_export
HejdaJakub: Yes, this is the way I wanted to do it in the first commit, but I did a mistake (I placed $values instead of $login). So I agree and I simplified it.

zlamalp (approved):

PR #570, opened 2021-02-25
Added new service to provision MyQ

rosinec:
- generates CSV import
- sends to a share

### DIFF
--- a/gen/myq_printsvc
+++ b/gen/myq_printsvc
@@ -1,0 +1,120 @@
+#!/usr/bin/perl                                                                                                                     
+
+use strict;
+use warnings;
+use perunServicesInit;
+use perunServicesUtils;
+use Text::CSV_XS;
+
+local $::SERVICE_NAME = "myq_printsvc";
+local $::PROTOCOL_VERSION = "3.1.0";
+my $SCRIPT_VERSION = "3.0.0";
+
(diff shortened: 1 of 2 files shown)


send/myq_printsvc
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+SERVICE_NAME="safeq"
```
zlamalp, line 3: `safeq` -> `myq_printsvc`

zlamalp (changes requested):

gen/myq_printsvc
```diff
@@ -0,0 +1,118 @@
...
+
+# GATHER USERS
+my $users;  # $users->{$login}->{ATTR} = $attrValue;
+
+# AGGREGATE DATA
+#
+foreach my $resourceId ( $data->getResourceIds() ){
+	# Iterate through resources assigned to the facility.
+
+	my $resourceName = $data->getResourceAttributeValue(resource => $resourceId, attrName => $A_R_NAME);
```
stavamichal, line 36: Is resource really represented in the external system by its name in the Perun? It should be an attribute on the resource IMO.

zlamalp:
Implementation is replaced with #626. Closing this one.

PR #571, opened 2021-03-01
Extend attributes with EPUID (identifier)

dBucik:
- added fetching of eduPersonUniqueIdentifier (urn:perun:user:attribute-def:def:login-namespace:bbmri-persistent-shadow)

### DIFF
--- a/gen/rt_bbmri
+++ b/gen/rt_bbmri
@@ -8,7 +8,7 @@ use JSON::XS;
 
 our $SERVICE_NAME = basename($0);
 our $PROTOCOL_VERSION = "3.1.1";
-my $SCRIPT_VERSION = "3.0.1";
+my $SCRIPT_VERSION = "3.0.2";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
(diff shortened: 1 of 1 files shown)


zlamalp (approved):

dBucik:
@zlamalp While testing, I have found out that for some reason the output for the added attribute is empty. Could that be because of using "def" version of the attribute instead of the "virt"?

zlamalp:
> @zlamalp While testing, I have found out that for some reason the output for the added attribute is empty. Could that be because of using "def" version of the attribute instead of the "virt"?

Well, only "virt" version of attribute generates the value if its empty and store it in "def" attribute. Question is, whether expected set of users send to this service is required to have it or not. I thought, the value is generated for all users, which goes through the registration process. If its not the case, you can make "virt" attribute required in this service and perun will generate it for all users assigned to this service.

dBucik:
> > @zlamalp While testing, I have found out that for some reason the output for the added attribute is empty. Could that be because of using "def" version of the attribute instead of the "virt"?
> 
> Well, only "virt" version of attribute generates the value if its empty and store it in "def" attribute. Question is, whether expected set of users send to this service is required to have it or not. I thought, the value is generated for all users, which goes through the registration process. If its not the case, you can make "virt" attribute required in this service and perun will generate it for all users assigned to this service.

Okay seems that it was just a delayed propagation result. Thanks for the explanation anyway!

PR #573, opened 2021-03-03
Finish statement after DBI calls in ad_user_mu_ucn

balcirakpeter:
- This script was executing two sql commands. One of them returned a
  value which was not readed. Therefore the statement pointer was not
  destroyed in the end of the script causing the process to result in
  error.
- For both statements was added chcek if the execution succeeded. Finish
  statement was added after the execution to make sure that the pointer is
  destroyed so the process will not result in error anymore.

### DIFF
--- a/send/ad_mu
+++ b/send/ad_mu
@@ -419,6 +419,7 @@ my $counter_add = 0;
 my $counter_updated = 0;
 my $counter_disabled = 0;
 my $counter_fail = 0;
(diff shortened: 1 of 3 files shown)


zlamalp (changes requested):
Should be ok, please fix this also in _ad_user_mu_ and _ad_mu_ services.

zlamalp (approved):

PR #576, opened 2021-03-09
Replacing deprecated method in du_users_export

cuadradek:
- du_users_export used method getRichUsersFromListOfUsersWithAttributes,
which is now deprecated and therefore it was replaced by new method
getRichUsersWithAttributesByIds.


zlamalp (approved):

stavamichal:
The same change needs to be done for service du_info_export (which is not in the git repository yet) manually.

PR #584, opened 2021-03-17
ad_mu changed to support cyclop workflow

balcirakpeter:
- Original changes can be found in PR https://github.com/CESNET/perun-services/pull/391
- This PR is rebased to master. Rebasing process was continualy done on
  test instance, so it is already tested.
- Gen script on test instance contained code to include "mandragora" in
  the email scopes. This was reverted in the final gen script.
- Gen script also contained "dev-" prefix in DNs. This was erased and
  original attribute for DNs is used again.
- Gen script also contained comments from Michal S. saying that for the
  adGroupDisplayName is used virt attribute instead of def. This comment
  was removed because the change is already used in production.
- Gen script also contained some format differences against master so it was
  reformated.
- Send script contained a change which forbided to call password_setter
  subroutine on the test instance. It was uncommented aggain as we want
  to call that method on production.
- Again format changes in send script.
- Send script in production had some parts which were commented
  (grace_periods). However, they were not commented in the send script
  from our master branch. This does not matter, because these parts with
  grace periods were completely erased from the final script.
- Gen was tested and it generates the same data as in the test instance,
  except the "mandragora" part in the scope of all emails and "dev-"
  prefix in DNs. (ad_mu.ldif was checked only partialy. It was not
  possible to vimdiff these two big files at once).
- Send script can be compiled but it was not tested.
- I advise to compare this two files against the files in the test
  instance not against the files in the master branch.

### DIFF
--- a/gen/ad_mu
+++ b/gen/ad_mu
@@ -7,7 +7,6 @@ use perunServicesUtils;
 use Time::Piece;
 
 sub processMembers;
-sub processExtendedLicenses;
 sub processGroups;
 
 local $::SERVICE_NAME = "ad_mu";
@@ -21,11 +20,7 @@ my $fileNameGroups = "$DIRECTORY/$::SERVICE_NAME"."_groups";  # printed for each
 my $fileNameOus = "$DIRECTORY/$::SERVICE_NAME"."_ous.ldif";
 my $baseDnFileName = "$DIRECTORY/baseDN";
 my $baseDnFileNameGroups = "$DIRECTORY/baseDNGroups";
-my $fileNameRelations = "$DIRECTORY/userRelations";
-my $fileNameLicenses = "$DIRECTORY/userLicenses";
-my $fileNameLicGroupNames = "$DIRECTORY/licGroupNames";
 my $fileNameFID = "$DIRECTORY/facilityId";
-my $fileNameManualEmployees = "$DIRECTORY/manualEmployees";
 
 my $data = perunServicesInit::getHashedDataWithGroups;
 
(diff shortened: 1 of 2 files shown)


gen/ad_mu
```diff
@@ -36,36 +31,37 @@ our $A_F_DOMAIN;  *A_F_DOMAIN = \'urn:perun:facility:attribute-def:def:adDomain'
...
 our $A_F_ID;  *A_F_ID = \'urn:perun:facility:attribute-def:core:id';
+our $A_F_O365_ALL_MAIN_LICENSES; *A_F_O365_ALL_MAIN_LICENSES = \'urn:perun:facility:attribute-def:def:o365AllMainLicenses';
+our $A_F_O365_USED_MAIN_LICENSES; *A_F_O365_USED_MAIN_LICENSES = \'urn:perun:facility:attribute-def:def:o365UsedMainLicenses';
 
 # OU, Group
 our $A_R_OU_NAME;  *A_R_OU_NAME = \'urn:perun:resource:attribute-def:def:adOuName';
-our $A_R_RELATION_TYPE;  *A_R_RELATION_TYPE = \'urn:perun:resource:attribute-def:def:relationType'; # STU/ZAM/ABS
 
-our $A_G_AD_NAME;    *A_G_AD_NAME = \'urn:perun:group:attribute-def:def:adName:o365mu';
+our $A_G_AD_NAME;  *A_G_AD_NAME = \'urn:perun:group:attribute-def:def:adName:o365mu';
```
stavamichal, line 40: In the commit message, you are telling that now the virtual group attribute for the name is used. But there is no change in the code.

gen/ad_mu
```diff
@@ -231,18 +208,16 @@ sub processGroups {
 			my $key = "CN=".$adName."_group.muni.cz,OU=".$ouName.",".$baseDNGroups;
 
 			$groups->{$ouName}->{$key}->{$A_G_AD_NAME} = $adName;
+			my $displayName = $data->getGroupAttributeValue( group => $groupId, attrName => $A_G_AD_DISPLAY_NAME );
```
stavamichal, line 211: This line of code already exists above.

stavamichal (changes requested):

gen/ad_mu
balcirakpeter: I was talking about display name -> A_G_AD_DISPLAY_NAME

gen/ad_mu
balcirakpeter: I missed this one. Probably forgoten line of code on the test instance. I have removed it.

gen/ad_mu
stavamichal: You are right, my bad.

stavamichal (approved):

PR #586, opened 2021-03-18
Added generic_send in Python

metodej:
- generic_send was rewritten in Python

### DIFF
--- a/send/generic_send.py
+++ b/send/generic_send.py
@@ -8,6 +8,7 @@ timeout_kill = "60"  # 60 sec to kill after timeout
 
 perun_cert = "/etc/perun/ssl/perun-send.pem"
 perun_key = "/etc/perun/ssl/perun-send.key"
+perun_chain = "/etc/perun/ssl/perun-send.chain"
 
 # predefined different types of destination
 destination_type_url = "url"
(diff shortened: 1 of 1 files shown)


send/generic_send.py
```diff
@@ -0,0 +1,177 @@
+import os
+import subprocess
+import sys
+import tempfile
+
+timeout = "5400"  # 90s * 60 sec = 1.5h
+timeout_kill = "60"  # 60 sec to kill after timeout
+
+perun_cert = "/etc/perun/ssl/perun-send.pem"
+perun_key = "/etc/perun/ssl/perun-send.key"
```
cuadradek, line 10: Please add:
`perun_chain="/etc/perun/ssl/perun-send.chain"`

send/generic_send.py
```diff
@@ -0,0 +1,177 @@
...
+	service_name = os.getenv('SERVICE_NAME')
+	if service_name is None:
+		print("Error: $SERVICE_NAME is not set", file=sys.stderr)
+		exit(1)
+
+	# choose transport command, only url type has different transport command at this moment
+	if destination_type == destination_type_url:
+		# add certificate to the curl if cert file and key file exists and they are readable
+		if os.access(perun_cert, os.R_OK) and os.access(perun_key, os.R_OK):
+			perun_cert_setting = "--cert " + perun_cert + " --key " + perun_key
```
cuadradek, line 57: Please, change this to:
```
if os.access(perun_cert, os.R_OK) and os.access(perun_key, os.R_OK) and os.access(perun_chain, os.R_OK):
	perun_cert_setting = "--cert " + perun_cert + " --key " + perun_key " --cacert " + perun_chain
```

cuadradek (changes requested):

martin-kuba:
This implementation is not more secure than the original in bash, because it still uses shell parsing. A secure implementation should create a new process by passing an array of arguments instead of a string.

cuadradek (approved):

PR #588, opened 2021-03-26
Skip users without proper licence in o365_mu

stavamichal:
- we want to skip all users by the same logic (not in the list of
 active o365 users - all not skipped users)
 - skip users if they have no active licence (licence is empty or 0) or
 if they have wrong UCO (not numerical format)
 - remove map of skipped users (not needed)
 - removed unused line in send script of o365


gen/o365_mu
```diff
@@ -342,9 +341,8 @@ sub processGroupMember {
 
 	my $UCO = $data->getUserFacilityAttributeValue( member => $memberId, attrName => $A_UF_LOGIN );
 	#skipped user which has no licence
-	if($skippedUsers->{$UCO}) {
-		return;
-	}
+	next unless( $users->{$UCO} );
```
balcirakpeter, line 344: return instead of next

balcirakpeter (changes requested):

gen/o365_mu
stavamichal: Right. Fixed.

balcirakpeter (approved):

PR #595, opened 2021-04-20
New service netbox

balcirakpeter:
- gen script fetches resources and their assigned members. For resources
  fetches netboxGroupName which serves as a group name in the destination
  and netboxIsStaff which indicates is_staff value for its members.
  Resources without the name attribute are skipped. We fetch name, email
  login, for each user. is_active is set to true and is_staff to false. These two
  are attributes in netbox. We set them in gen script in case, we will
  fill them from perun in the future. groups list is filled with
  netboxGroupName attribute for each member's resource.
- send script creates missing groups first. Then creates missing users.
  In this part we generate also password and set group membership. In
  the end we update users. If user does not exists in perun but exists
  in netbox he is set to inactive and his group list is erased. If he
  exists both in perun and netbox, his attributes are updated except
  password. laso group membership is updated in this part. We do not
  delete users and groups. When group does not exists in perun but
  exists in netbox. It will be empty in the end of the process.
- for this service is needed file with name netbox in /etc/perun/, which
  contains authorization token. Destination needs to be set to without
  trailing "/".

### DIFF
--- a/send/netbox
+++ b/send/netbox
@@ -1,0 +1,262 @@
+#!/usr/bin/perl
+use strict;
+use warnings;
+
+use Getopt::Long qw(:config no_ignore_case);
+use LWP::UserAgent;
+use HTTP::Request;
+use HTTP::Headers;
+use JSON;
+use POSIX qw(strftime);
+use Encode qw(decode_utf8);
+
+#We want to have data in UTF8 on output
+binmode STDOUT, ':utf8';
+binmode STDERR, ':utf8';
+
+#parse arguments in utf8
+@ARGV = map { decode_utf8($_, 1) } @ARGV;
+
+# define service
+my $service_name = "netbox";
+
+# facility name
+my $facility_name = $ARGV[0];
+chomp($facility_name);
+
+# propagation destination
+my $destination = $ARGV[1];
+chomp($destination);
+
+# GEN folder location
+my $service_files_base_dir="../gen/spool";
+my $service_files_dir="$service_files_base_dir/$facility_name/$service_name";
+my $users_file_name = "$service_files_dir/netbox_users";
+my $groups_file_name = "$service_files_dir/netbox_groups";
+
+# Authorization TOKEN
+open(my $file, '<', "/etc/perun/$service_name")
+	or die("Can't open /etc/perun/$service_name: $!\n");
+my $TOKEN = <$file>;
+chomp($TOKEN);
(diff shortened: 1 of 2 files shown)


send/netbox
```diff
@@ -0,0 +1,250 @@
...
+# callServer
+#-----------------------
+#Parameters:
+# url     - url of server and method address to call
+# type    - GET, POST or PATCH
+# content - hash content of request, will be encoded to json output
+#-----------------------
+#Returns:
+# resolved status of call
+# 0 means everything is ok and done
```
cuadradek, line 155: I would change this to something like: Returns hash of response content

send/netbox
```diff
@@ -0,0 +1,250 @@
...
+	return check_server_response( $server_response );
+}
+
+#Name:
+# createConnection
+#-----------------------
+#Parameters:
+# url     - url of server and method address to call
+# type    - GET, POST or PATCH
+# content - hash content of request, will be encoded to json output
```
cuadradek, line 180: content is already passed as json

send/netbox
```diff
@@ -0,0 +1,250 @@
...
+#-----------------------
+sub create_connection {
+	my $url = shift;
+	my $type = shift;
+	my $content = shift;
+
+	my $headers = HTTP::Headers->new;
+	$headers->header('Content-Type' => 'application/json');
+	$headers->header('Accept' => 'application/json,text/json');
+	$headers->header('Authorization' => 'Token ' . $TOKEN);                                                       #ASI DOBRE
```
cuadradek, line 196: You probably forgot to delete the comment

cuadradek (changes requested):
Looks good to me, just a couple of unclarities in comments

send/netbox
balcirakpeter: Fixed

send/netbox
balcirakpeter: Fiked

send/netbox
balcirakpeter: Fixed

cuadradek (approved):

rosinec:
Provisioning scripts are well written and meets all requirements that were desired by service provider. 
Provisioning process was tested with the service provider.

rosinec (approved):
Provisioning scripts are well written and meets all requirements that were desired by service provider.
Provisioning process was tested with the service provider.

PR #597, opened 2021-04-30
Updated slave scripts for fs_home and k5login

cuadradek:
- In process-fs_home it is now possible to skip creating a home
directory from mid hook.
- process-k5login now skips creating k5login files in home directories
that don't exist.

### DIFF
--- a/slave/process-k5login/changelog
+++ b/slave/process-k5login/changelog
@@ -1,3 +1,9 @@
+perun-slave-process-k5login (3.1.7) stable; urgency=low
+
+  * Skip creating k5login files in home directories that don't exist.
+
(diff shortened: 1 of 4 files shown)


slave/process-k5login/changelog
```diff
@@ -1,3 +1,9 @@
+perun-slave-process-k5login (3.1.6) stable; urgency=low
```
Vojtech-Sassmann, line 1: Shouldn't the patch version be updated?
```suggestion
perun-slave-process-k5login (3.1.7) stable; urgency=low
```

cuadradek: Fixed, thanks.

balcirakpeter (approved):

PR #601, opened 2021-05-27
feat: new script ad_group_vsup_o365

Vojtech-Sassmann:
* Created new script ad_group_vsup_o365 which respects vsup's expiration.

### DIFF
--- a/gen/ad_group_vsup_o365
+++ b/gen/ad_group_vsup_o365
@@ -1,0 +1,161 @@
+#!/usr/bin/perl
+use feature "switch";
+use strict;
+use warnings;
+use perunServicesInit;
+use perunServicesUtils;
+
+local $::SERVICE_NAME = "ad_group_vsup_o365";
+local $::PROTOCOL_VERSION = "3.0.0";
+my $SCRIPT_VERSION = "3.0.0";
+
+perunServicesInit::init;
+my $DIRECTORY = perunServicesInit::getDirectory;
+my $fileName = "$DIRECTORY/$::SERVICE_NAME".".ldif";
+my $baseDnFileName = "$DIRECTORY/baseDN";
+
+my $data = perunServicesInit::getHashedHierarchicalData;
+
+#Constants
+our $A_LOGIN; *A_LOGIN = \'urn:perun:user_facility:attribute-def:virt:login';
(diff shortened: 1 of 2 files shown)


gen/ad_group_vsup_o365
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	# Add time 23:59:59 to the date, since we want accounts to be active on the last day
+	$latest_expiration = $latest_expiration + 86399;
+
+	if ($latest_expiration > $currentDate->epoch) {
+		return 'active';
+	}
+
+	# (will) expire by studies - add 14 days grace period
```
zlamalp, line 161: This service should have only +14 days grace period, so disregarding users relation (employee/student) - meaning expiration source - you can always check only latest expiration + 14 days against the current date. Also then you can probably return simple 0/1 from this method.

zlamalp (changes requested):

gen/ad_group_vsup_o365
```diff
@@ -0,0 +1,183 @@
...
+
+}
+
+close FILE;
+
+perunServicesInit::finalize;
+
+
+#
+# active - výchozí stav, netřeba nastavovat při vytváření schránky, pouze pokud byl účet předtím v jiném stavu
```
balcirakpeter, line 123: NIT: It would be better to use English in the comments :)

balcirakpeter (approved):

gen/ad_group_vsup_o365
```diff
@@ -0,0 +1,165 @@
...
+
+	if (!defined $expirationKos and !defined $expirationDc2 and !defined $expirationMan) {
+		# if no expiration set in source data - take as "never"
+		return 1;
+	}
+
+	# Add time 23:59:59 to the date, since we want accounts to be active on the last day
+	$latest_expiration = $latest_expiration + 86399;
+
+	if ($latest_expiration > $currentDate->epoch) {
```
zlamalp, line 156: Since we don't distinguish valid users from those in grace period, you can probably remove this condition and use only the second, which adds 14 days to expiration date.

zlamalp (changes requested):

gen/ad_group_vsup_o365
Vojtech-Sassmann: Ok, done.

zlamalp (approved):
Seems to be OK. I'll test it on production next week, but it can be merged.

PR #602, opened 2021-06-02
Edited service mailaliases

HejdaJakub:
* This service now allows only valid members by default.
* If is set new attribute on resource, then are allowed also expired members.

### DIFF
--- a/gen/mailaliases
+++ b/gen/mailaliases
@@ -7,7 +7,7 @@ use perunServicesUtils;
 
 local $::SERVICE_NAME = "mailaliases";
 local $::PROTOCOL_VERSION = "3.0.0";
-my $SCRIPT_VERSION = "3.0.2";
+my $SCRIPT_VERSION = "3.0.3";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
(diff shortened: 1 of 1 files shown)


gen/mailaliases
```diff
@@ -16,17 +16,33 @@ my $data = perunServicesInit::getHashedHierarchicalData;
...
+		if($allowExpiredMembers) {
+			if(($status ne $STATUS_VALID) && ($status ne $STATUS_EXPIRED)) { next; }
+		} else {
+			if($status ne $STATUS_VALID) { next; }
+		}
 
-	$mailByLogin{$login} = $mail;
+		my $login  = $data->getUserFacilityAttributeValue( member => $memberId, attrName => $A_USER_LOGIN);
+		my $mail  = $data->getUserAttributeValue( member => $memberId, attrName => $A_USER_MAIL);
+		$mailByLogin{$login} = $mail;
```
balcirakpeter, line 44: The code seems to be ok, I would just simplify this part so it is more readable.
Something like this:
```
if (($status eq $STATUS_VALID) || ($status eq $STATUS_EXPIRED && $allowExpiredMembers)) {
    my $login  = $data->getUserFacilityAttributeValue( member => $memberId, attrName => $A_USER_LOGIN);		 
    my $mail  = $data->getUserAttributeValue( member => $memberId, attrName => $A_USER_MAIL);		 
    $mailByLogin{$login} = $mail;
}
```

balcirakpeter (changes requested):

gen/mailaliases
HejdaJakub: It looks good! I rewrote it.

balcirakpeter (approved):

PR #606, opened 2021-06-18
Hashed data - denbi*, docdb, drupal*

Johaney-s:
* denbi_portal_compute_center, docdb, drupal_elixir now using hashedData

### DIFF
--- a/gen/denbi_portal_compute_center
+++ b/gen/denbi_portal_compute_center
@@ -12,11 +12,11 @@ use utf8;
 
 local $::SERVICE_NAME = "denbi_portal_compute_center";
 local $::PROTOCOL_VERSION = "1.0.0";
-my $SCRIPT_VERSION = "1.0.0";
+my $SCRIPT_VERSION = "1.0.1";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
-my $data = perunServicesInit::getDataWithGroups;
+my $data = perunServicesInit::getHashedDataWithGroups;
 
 #forward declaration
 sub processUsers;
@@ -77,10 +77,10 @@ my $fileGroups = $DIRECTORY . "/groups.scim";
 
 my $agent = perunServicesInit->getAgent;
 
-foreach my $resourceData ($data->getChildElements) {
-        foreach my $groupData (($resourceData->getChildElements)[0]->getChildElements) {
-                my $groupMembersLogins = processGroups $groupData;
-        }
+foreach my $resourceId ($data->getResourceIds()) {
+	foreach my $groupId ($data->getGroupIdsForResource( resource => $resourceId )) {
+		processGroups($groupId, $resourceId);
+	}
 }
 
 # PREPARE USERSDATA TO JSON
@@ -99,7 +99,7 @@ my @users = values %$userStruc;
 # PRINT USERS TO JSON
 open FILE_USERS,">$fileUsers" or die "Cannot open $fileUsers: $! \n";
 binmode(FILE_USERS);
-print FILE_USERS JSON::XS->new->utf8->pretty->encode(\@users);
+print FILE_USERS JSON::XS->new->utf8->pretty->canonical->encode(\@users);
 close (FILE_USERS) or die "Cannot close $fileUsers: $! \n";
 
 # PREPARE GROUPSDATA TO JSON
@@ -124,7 +124,7 @@ my @groups = values %$groupStruc;
 # PRINT GROUPS TO JSON
 open FILE_GROUPS,">$fileGroups" or die "Cannot open $fileGroups: $! \n";
 binmode(FILE_GROUPS);
-print FILE_GROUPS JSON::XS->new->utf8->pretty->encode(\@groups);
+print FILE_GROUPS JSON::XS->new->utf8->pretty->canonical->encode(\@groups);
 close (FILE_GROUPS) or die "Cannot close $fileGroups: $! \n";
 
 perunServicesInit::finalize;
@@ -140,12 +140,22 @@ sub convertToFriendlyName($) {
 
 ## creates structure for users.scim file
 sub processUsers {
-        my ($gid, $memberData) = @_;
+	my ($gid, $memberId, $resourceId) = @_;
 
-        my %memberAttributes = attributesToHash $memberData->getAttributes;
-        my $uid = $memberAttributes{$A_USER_ID};
-        my $status = $memberAttributes{$A_USER_STATUS};
-	
+	my $userTypeAttributes = $data->getAllUserAttributes(member => $memberId);
+	my $memberTypeAttributes = $data->getAllMemberAttributes(member => $memberId);
+	my $memberResourceTypeAttributes = $data->getAllMemberResourceAttributes(member => $memberId, resource => $resourceId);
+	my $userFacilityTypeAttributes = $data->getAllUserFacilityAttributes(member => $memberId, facility => $data->getFacilityId());
+	my $memberGroupTypeAttributes = $data->getAllMemberGroupAttributes(member => $memberId, group => $gid);
+
+	my %memberAttributes = ($userTypeAttributes ? %{$userTypeAttributes} : (),
+							$memberTypeAttributes ? %{$memberTypeAttributes} : (),
+							$memberResourceTypeAttributes ? %{$memberResourceTypeAttributes} : (),
+							$userFacilityTypeAttributes ? %{$userFacilityTypeAttributes} : (),
+							$memberGroupTypeAttributes ? %{$memberGroupTypeAttributes} : ());
+	my $uid = $data->getUserAttributeValue( member => $memberId, attrName => $A_USER_ID );
+	my $status = $data->getMemberAttributeValue( member => $memberId, attrName => $A_USER_STATUS );
+
 	if($status ne $STATUS_VALID) { return; }
 
 	unless($userStruc->{$uid}) {
@@ -153,14 +163,13 @@ sub processUsers {
 		$userStruc->{$uid} = \%memberAttributes;
 	}
 
-        processMemberships $gid, $uid;
+	processMemberships $gid, $uid;
 }
 
 ## creates structure for groups.scim file
 sub processGroups {
-        my $group = shift;
-        my %groupAttributes = attributesToHash $group->getAttributes;
-        my $membersElement = ($group->getChildElements)[1];
+	my ($groupId, $resourceId) = @_;
+	my %groupAttributes= %{$data->getAllGroupAttributes(group => $groupId)};
 
 	my $gid = $groupAttributes{$A_GROUP_ID};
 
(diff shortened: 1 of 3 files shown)


Johaney-s:
Tested all scripts; please check denbi* if it can use the attributes the way I propose (explicitly naming them) or there is some simpler way to do it. Not sure some attributes won't be missing if more attributes would be available for this service...

cuadradek (approved):
Looks good, just one minor thing - I suggest using `->canonical` when printing JSON, the JSON is then sorted and outputs of scripts are easily comparable:

`print FILE JSON::XS->new->utf8->pretty->canonical->encode(\@users);`

Johaney-s:
> Looks good, just one minor thing - I suggest using `->canonical` when printing JSON, the JSON is then sorted and outputs of scripts are easily comparable:
> 
> `print FILE JSON::XS->new->utf8->pretty->canonical->encode(\@users);`

Thanks, added `canonical`.

gen/denbi_portal_compute_center
```diff
@@ -140,48 +132,49 @@ sub convertToFriendlyName($) {
...
-
-	my $gid = $groupAttributes{$A_GROUP_ID};
-
-	#convert attributes data
-	$groupAttributes{$A_GROUP_DENBI_PROJECT_MEMBERS} = decode_json $groupAttributes{$A_GROUP_DENBI_PROJECT_MEMBERS};
-
-	unless(exists $groupStruc->{$gid}) {
-		%groupAttributes = map { convertToFriendlyName($_) => $groupAttributes{$_} } keys %groupAttributes;
-		$groupStruc->{$gid} = \%groupAttributes;
+        my $groupId = shift;
```
balcirakpeter, line 156: Wrong indentation :)

gen/denbi_portal_compute_center
```diff
@@ -140,48 +132,49 @@ sub convertToFriendlyName($) {
...
-        my %groupAttributes = attributesToHash $group->getAttributes;
-        my $membersElement = ($group->getChildElements)[1];
-
-	my $gid = $groupAttributes{$A_GROUP_ID};
-
-	#convert attributes data
-	$groupAttributes{$A_GROUP_DENBI_PROJECT_MEMBERS} = decode_json $groupAttributes{$A_GROUP_DENBI_PROJECT_MEMBERS};
-
-	unless(exists $groupStruc->{$gid}) {
-		%groupAttributes = map { convertToFriendlyName($_) => $groupAttributes{$_} } keys %groupAttributes;
```
balcirakpeter, line 171: We have to check if the service used only defined attributes, which are defined at the beginning of the script. It seems that it could have generated also other attributes (commented at the beginning) as it iterated through all of the attributes on this line, not only the defined ones. @dBucik could you help us with that?

gen/denbi_portal_compute_center
```diff
@@ -140,48 +132,49 @@ sub convertToFriendlyName($) {
 
 ## creates structure for users.scim file
 sub processUsers {
-        my ($gid, $memberData) = @_;
+		my $groupId = shift;
+		my $memberId = shift;
+
+        my $userId = $data->getUserAttributeValue( member => $memberId, attrName => $A_USER_ID );
```
balcirakpeter, line 138: Wrong indentation :)

gen/denbi_portal_compute_center
```diff
@@ -140,48 +132,49 @@ sub convertToFriendlyName($) {
...
+		my $status = $data->getMemberAttributeValue( member => $memberId, attrName => $A_USER_STATUS );
 
-        my %memberAttributes = attributesToHash $memberData->getAttributes;
-        my $uid = $memberAttributes{$A_USER_ID};
-        my $status = $memberAttributes{$A_USER_STATUS};
 	
 	if($status ne $STATUS_VALID) { return; }
 
-	unless($userStruc->{$uid}) {
-		%memberAttributes = map { convertToFriendlyName($_) => $memberAttributes{$_} } keys %memberAttributes;
```
balcirakpeter, line 152: @dBucik same question as for the group attributes.

balcirakpeter (changes requested):

gen/denbi_portal_compute_center
Johaney-s: Fixed, used methods from https://github.com/CESNET/perun/pull/3264

gen/denbi_portal_compute_center
```diff
@@ -99,7 +99,7 @@ my @users = values %$userStruc;
 # PRINT USERS TO JSON
 open FILE_USERS,">$fileUsers" or die "Cannot open $fileUsers: $! \n";
 binmode(FILE_USERS);
-print FILE_USERS JSON::XS->new->utf8->pretty->encode(\@users);
+print FILE_USERS JSON::XS->new->utf8->pretty->canonical->encode(\@users);
```
zlamalp, line 102: Resulting `users.scim` file is missing `groupMembershipExpiration` property of JSON objects for new script version while old version has it. I'm not sure where is the problem yet.

zlamalp (changes requested):

gen/docdb
```diff
@@ -36,11 +36,11 @@ my $outputData = {
 						 members => [],
 					 };
 
-foreach my $resourceData ($data->getChildElements) {
-	processResourceData $resourceData;
+foreach my $resourceId ($data->getResourceIds()) {
```
zlamalp, line 39: `our $A_RESOURCE_VO_NAME;` is no longer used and that line can be removed.

gen/drupal_elixir
```diff
@@ -67,9 +67,9 @@ my $fileMemberships = $DIRECTORY . "memberships.csv";
 my $agent = perunServicesInit->getAgent;
 my $groupAgent = $agent->getGroupsAgent;
 
-foreach my $resourceData ($data->getChildElements) {
-	foreach my $groupData (($resourceData->getChildElements)[0]->getChildElements) {
-		my $groupMembersLogins = processGroups $groupData;
+foreach my $resourceId ( $data->getResourceIds() ) {
+	foreach my $groupId ( $data->getGroupIdsForResource( resource => $resourceId ) ) {
+		my $groupMembersLogins = processGroups $groupId, $resourceId;
```
zlamalp, line 72: Variable `$groupMembersLogins` is not used, you can simply call `processGroups()` method without processing returned value.

gen/docdb
```diff
@@ -36,11 +36,11 @@ my $outputData = {
...
+	processResourceMembersData $resourceId;
 
-	foreach my $groupData (($resourceData->getChildElements)[0]->getChildElements) {
-		my $groupMembersLogins = processGroupData $groupData;
+	foreach my $groupId ($data->getGroupIdsForResource( resource => $resourceId )) {
+		my $groupMembersLogins = processGroupData $resourceId, $groupId;
```
zlamalp, line 43: Variable `$groupMembersLogins` is not used, you can simply call `processGroups()` method without processing returned value.

gen/docdb
```diff
@@ -49,56 +49,46 @@ foreach my $resourceData ($data->getChildElements) {
...
 sub processGroupsMembersData {
-	my ($membersElement, $groupName) = @_;
+	my ($resourceId, $groupId) = @_;
 	my @groupMembers = ();
-	for my $memberData ($membersElement->getChildElements) {
-		my %memberAttributes = attributesToHash $memberData->getAttributes;
-		push @groupMembers, $memberAttributes{$A_USER_LOGIN};
+	foreach my $memberId($data->getMemberIdsForResourceAndGroup( resource => $resourceId, group => $groupId )) {
+		push @groupMembers, $data->getUserFacilityAttributeValue( member => $memberId, attrName => $A_USER_LOGIN );
 	}
```
zlamalp, line 92: You can sort `@groupMembers` like this:
```perl
@groupMembers = sort @groupMembers;
```
before returning them. It will create consistent order in output. Since it contains only users logins, their order shouldn't affect the service, but it will enable us to have consistent and comparable data.

gen/drupal_elixir
```diff
@@ -67,9 +67,9 @@ my $fileMemberships = $DIRECTORY . "memberships.csv";
 my $agent = perunServicesInit->getAgent;
```
zlamalp, line 67: `$usedEmailsStruc` seems to be unused.

zlamalp (changes requested):

gen/denbi_portal_compute_center
Johaney-s: Original member data contained user, member, member-resource, user-facility and member-group attributes, added all.

gen/docdb
Johaney-s: removed

gen/drupal_elixir
Johaney-s: removed variable

gen/docdb
Johaney-s: added

gen/drupal_elixir
Johaney-s: removed

gen/drupal_elixir
```diff
@@ -175,42 +175,39 @@ sub processUsers {
...
+
+	if($data->getGroupAttributeValue( group => $groupId, attrName => $A_GROUP_NAME )) {
+		my $groupName = $data->getGroupAttributeValue( group => $groupId, attrName => $A_GROUP_NAME ),
+		my $groupParentID = $data->getGroupAttributeValue( group => $groupId, attrName => $A_PARENT_GROUP_ID );
 		my $groupType;
 
-		if($groupAttributes{$A_GROUP_TYPE}) {
-			$groupType = $groupAttributes{$A_GROUP_TYPE};
+		if($data->getGroupResourceAttributeValue( group => $groupId, resource=> $resourceId, attrName => $A_GROUP_TYPE )) {
+			$groupType = $data->getGroupResourceAttributeValue( group => $groupId, attrName => $A_GROUP_TYPE );
```
zlamalp, line 196: Since attribute is group-resource, you must pass also `resource => $resourceId` param to the method.

gen/drupal_elixir
```diff
@@ -67,9 +66,9 @@ my $fileMemberships = $DIRECTORY . "memberships.csv";
 my $agent = perunServicesInit->getAgent;
 my $groupAgent = $agent->getGroupsAgent;
 
-foreach my $resourceData ($data->getChildElements) {
-	foreach my $groupData (($resourceData->getChildElements)[0]->getChildElements) {
-		my $groupMembersLogins = processGroups $groupData;
+foreach my $resourceId ( $data->getResourceIds() ) {
```
zlamalp, line 69: Because entities are iterated over in ordered way now, user entries in `users-duplicities.csv` and `users.csv` has changed. It's because drupal (end service) considers same email as the same user. Since `groups.csv` and `memberships.csv` hasn't changed, I think it shouldn't affect the users -> they should have access to the same resources in the drupal. What do you think @stavamichal ?

zlamalp (changes requested):

gen/drupal_elixir
zlamalp: I discussed this with service owner and this change is OK for them, since they then manually adjust users in drupal instance.

gen/drupal_elixir
Johaney-s: Fixed

balcirakpeter (approved):

zlamalp (approved):

PR #607, opened 2021-06-22
Hashed data - openvpn, operations_portal_egi, owncloud_vo_mapping

Johaney-s:
* Using hashed data in these services
* All scripts tested

### DIFF
--- a/gen/operations_portal_egi
+++ b/gen/operations_portal_egi
@@ -7,10 +7,11 @@ use File::Basename;
 
 our $SERVICE_NAME = basename($0);
 our $PROTOCOL_VERSION = "3.0.0";
+our $SCRIPT_VERSION = "3.0.0";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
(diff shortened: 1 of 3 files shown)


metodej (approved):

gen/operations_portal_egi
```diff
@@ -6,11 +6,11 @@ use perunServicesUtils;
 use File::Basename;
 
 our $SERVICE_NAME = basename($0);
-our $PROTOCOL_VERSION = "3.0.0";
+our $PROTOCOL_VERSION = "3.0.1";
```
zlamalp, line 9: Careful, this is `$PROTOCOL_VERSION` and not `$SCRIPT_VERSION`. You haven't changed the protocol as script provides same data structure. Script seems to be missing `$SCRIPT_VERSION` so you can add it and set it to `3.0.0`.

zlamalp (changes requested):

gen/operations_portal_egi
Johaney-s: I see, thanks. Fixed.

zlamalp (approved):

PR #611, opened 2021-07-01
Hashed data - samba_du, scim, sympa*

Johaney-s:
* these services now use hashedData

### DIFF
--- a/gen/scim
+++ b/gen/scim
@@ -10,11 +10,11 @@ use utf8;
 
 local $::SERVICE_NAME = "scim";
 local $::PROTOCOL_VERSION = "1.0.0";
-my $SCRIPT_VERSION = "1.0.0";
+my $SCRIPT_VERSION = "1.0.1";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
-my $data = perunServicesInit::getDataWithGroups;
+my $data = perunServicesInit::getHashedDataWithGroups;
 
 #forward declaration
 sub processUsers;
@@ -48,10 +48,10 @@ my $userLogin = {};
 my $fileUsers = $DIRECTORY . "/users.scim";
 my $fileGroups = $DIRECTORY . "/groups.scim";
 
-foreach my $resourceData ($data->getChildElements) {
-        foreach my $groupData (($resourceData->getChildElements)[0]->getChildElements) {
-                processGroups $groupData;
-        }
+foreach my $resourceId ($data->getResourceIds()) {
+	foreach my $groupId ($data->getGroupIdsForResource( resource => $resourceId )) {
+		processGroups $groupId, $resourceId;
+    }
 }
 
 # PREPARE USERSDATA TO JSON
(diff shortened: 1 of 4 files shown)


gen/sympa
```diff
@@ -57,7 +51,7 @@ for my $listName (keys %$mailinglistStruc) {
     open FILE,">$fileName" or die "Cannot open $fileName: $! \n";
 
     for my $mail (keys %{$mailinglistStruc->{$listName}}) {
```
metodej, line 53: I think we could add `sort` here to have unified output just like in the `symba_cesnet`.

gen/scim
```diff
@@ -70,16 +70,16 @@ foreach my $uid (sort keys %$userStruc) {
...
 
 # PREPARE GROUPSDATA TO JSON
 my @groups;
 foreach my $gid (sort keys %$groupStruc) {
         my $group = {};
         $group->{"id"} = $gid;
-        $group->{"name"} = $groupStruc->{$gid}->{$A_GROUP_NAME};
-        $group->{"parentGroupId"} = $groupStruc->{$gid}->{$A_GROUP_PAR_ID};
+        $group->{"name"} = $data->getGroupAttributeValue(group => $gid, attrName => $A_GROUP_NAME);
+        $group->{"parentGroupId"} = $data->getGroupAttributeValue(group => $gid, attrName => $A_GROUP_PAR_ID);
```
metodej, line 82: I think you can get the attributes from the `groupStruc` just like it used to do or you don't have save it in there during `processGroups` method.

metodej (changes requested):
The scripts look great, I have just some comments.

gen/scim
```diff
@@ -140,22 +139,20 @@ sub processUsers {
...
-
-        if ($groupAttributes{$A_GROUP_NAME}) {
-                my $groupName = $groupAttributes{$A_GROUP_NAME};
-                my $gid = $groupAttributes{$A_GROUP_ID};
-                my $groupParId = $groupAttributes{$A_GROUP_PAR_ID};
-
-                unless(exists $groupStruc->{$gid}) {
-                        $groupStruc->{$gid}->{$A_GROUP_NAME} = $groupName;
-                        $groupStruc->{$gid}->{$A_GROUP_PAR_ID} = $groupParId;
+        my $groupId = shift;
```
balcirakpeter, line 142: bad indentation

balcirakpeter (approved):
The code seems to be ok. Just a small thing. It would be better to be consistent in parameter ordering while calling `getAttributeValues` subs. It will be more readable.

gen/scim
Johaney-s: fixed

gen/scim
Johaney-s: You're right, kept the old approach.

gen/sympa
Johaney-s: added

metodej (approved):

gen/sympa
```diff
@@ -25,29 +25,23 @@ our $A_USER_STATUS;                      *A_USER_STATUS =                     \'
...
-
-    my @membersData = $rData->getChildElements;
+foreach my $resourceId ($data->getResourceIds()) {
+	my $listName = $data->getResourceAttributeValue(resource => $resourceId, attrName => $A_RESOURCE_MAILING_LIST_NAME);
 
-    my $listName = "$resourceAttributes{$A_RESOURCE_MAILING_LIST_NAME}";
-
-    foreach my $memberAttributes (dataToAttributesHashes @membersData) {
+	for my $memberId ($data->getMemberIdsForResource(resource => $resourceId)) {
         #list only VALID members except if allowNonvalidUsers is true
```
zlamalp, line 38: I'ts not your change, but this comment doesn't make sense and its probably copy-paste from original mailman script. We do not check any resource attribute to allow non-valid users, script forces usage of only valid users by the next command. Please remove the comment. Same goes for `sympa_cesnet` version of the service.

zlamalp (changes requested):

gen/sympa
Johaney-s: Removed.

zlamalp (approved):

PR #612, opened 2021-07-02
perf(gen): Hashed data - pkinit, rt*, safeq*

cuadradek:
Services pkinit, rt, rt_bbmri, safeq and safeq5 are now using
getHashedHierarchicalData.

### DIFF
--- a/gen/safeq
+++ b/gen/safeq
@@ -18,7 +18,7 @@ local $::PROTOCOL_VERSION = "3.0.0";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
-my $data = perunServicesInit::getHierarchicalData;
+my $data = perunServicesInit::getHashedHierarchicalData;
 
 #Constants
 our $A_F_BASE_DN;              *A_F_BASE_DN =             \'urn:perun:facility:attribute-def:def:adBaseDN';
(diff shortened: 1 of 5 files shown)


Johaney-s (approved):
looks good

metodej (approved):
It looks great, I have just a nitpick I think it would be nice to be consistent in parameters ordering in `get*AttributeValue` methods...

gen/safeq
```diff
@@ -141,25 +136,32 @@ perunServicesInit::finalize;
...
-	my $memberAttributes = shift;
-	my $resourceAttributes = shift;
-	my $login = $memberAttributes->{$A_USER_LOGIN};
-
-	unless(defined $usersByLogin->{$memberAttributes->{$A_USER_LOGIN}}) {
-	my %tmp =  ( %$memberAttributes, %$resourceAttributes );
-		$usersByLogin->{$memberAttributes->{$A_USER_LOGIN}} = $memberAttributes;
-		$usersByLogin->{$memberAttributes->{$A_USER_LOGIN}}->{$A_R_PREPAID} = $resourceAttributes->{$A_R_PREPAID};
-		$usersByLogin->{$memberAttributes->{$A_USER_LOGIN}}->{$A_R_PRINT_FOR_FREE} = ( $resourceAttributes->{$A_R_PRINT_FOR_FREE} || 0) ;
+	my $memberId = shift;
```
zlamalp, line 139: See where you print `role:` in LDIF file. You can probably use `sort keys ...`  to get comparable outputs. It's same for `billingcode:` or `card:` fields.

zlamalp (changes requested):

gen/safeq
zlamalp: I'm waiting for confirmation from service owner that we can sort those attributes, so please wait with that cahnge. Also it seems, that `safeq5` version can be deleted instead - also waiting for confirmation, so we won't merge this yet.

PR #615, opened 2021-08-02
refactor(send): removed locks from ad_mu service

balcirakpeter:
Service ad_mu used locks to prevent starting after the previous run
failed. It was used because the service worked with cache files which
could contain wrong data, so the next run could compute incorect results.
However, cached files are not used anymore, so the locks are
pointless. Therefore, they were removed from the script.

### DIFF
--- a/send/ad_mu
+++ b/send/ad_mu
@@ -33,7 +33,6 @@ use Net::LDAP::Control::Paged;
 use Net::LDAP::Constant qw( LDAP_CONTROL_PAGED );
 use File::Copy;
 use Date::Calc qw/ Today Delta_Days Add_Delta_Days Date_to_Days Date_to_Text Decode_Date_EU /;
-use LockFile::Simple;
 no if $] >= 5.017011, warnings => 'experimental::smartmatch';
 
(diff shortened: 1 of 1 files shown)


send/ad_mu
```diff
@@ -95,16 +93,6 @@ close $file_g;
 my $namespace = $ARGV[1];
 chomp($namespace);
 
-# create service lock
-my $lock = ScriptLock->new($facility_name . "_" . $service_name . "_" . $namespace);
-($lock->lock() == 1) or die "Unable to get lock, service propagation is already running or previous run failed.";
```
zlamalp, line 100: I think you should keep our standard lock `ScriptLock` as it prevents running the script concurrently and has no effect on failed status of the script (error lock which can be removed). See eg. `ad_group_vsup` service script.

zlamalp (changes requested):

send/ad_mu
balcirakpeter: Fixed

zlamalp (approved):

PR #616, opened 2021-08-02
feat(gen): skipping inactive members in ad_group_mu_ucn

Johaney-s:
* Other than valid members are skipped, expired members can be propagated if relevant attribute is set

### DIFF
--- a/gen/ad_group_mu_ucn
+++ b/gen/ad_group_mu_ucn
@@ -7,7 +7,7 @@ use perunServicesUtils;
 
 local $::SERVICE_NAME = "ad_group_mu_ucn";
 local $::PROTOCOL_VERSION = "3.0.0";
-my $SCRIPT_VERSION = "3.0.1";
+my $SCRIPT_VERSION = "3.0.2";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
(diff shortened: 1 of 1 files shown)


Johaney-s:
`getHashedHierarchicalData(1)` filters based on group status, `memberStatus`based on VO status, so they don't affect each other

gen/ad_group_mu_ucn
```diff
@@ -56,10 +62,13 @@ foreach my $resourceId ($data->getResourceIds()) {
 
 		my $login = $data->getUserFacilityAttributeValue( member => $memberId, attrName => $A_LOGIN );
 		my $isBanned = $data->getMemberResourceAttributeValue( member => $memberId, resource => $resourceId, attrName => $A_MR_V_IS_BANNED );
+		my $memberStatus = $data->getMemberAttributeValue( attrName => $A_MEMBER_STATUS, member => $memberId );
```
metodej, line 65: I think the code looks okay overall, just a nitpick about the argument order.

metodej (approved):

gen/ad_group_mu_ucn
Johaney-s: Thanks, fixed the ordering.

balcirakpeter (approved):

PR #618, opened 2021-08-11
feat: kerberos_renewal_principals service

Johaney-s:
* This service returns all kerberos principals from valid users

### DIFF
--- a/gen/kerberos_renewal_principals
+++ b/gen/kerberos_renewal_principals
@@ -1,0 +1,42 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use perunServicesInit;
+use perunServicesUtils;
+use File::Basename;
+
+local $::SERVICE_NAME = basename($0);
+local $::PROTOCOL_VERSION = "3.0.0";
+my $SCRIPT_VERSION = "3.0.0";
+
+perunServicesInit::init;
+my $data = perunServicesInit::getHashedHierarchicalData;
+my $DIRECTORY = perunServicesInit::getDirectory;
+
+
+our $A_PRINCIPALS;                    *A_PRINCIPALS =                       \'urn:perun:user:attribute-def:def:kerberosLogins';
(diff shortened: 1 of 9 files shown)


metodej (approved):
The code looks good!

gen/kerberos_renewal_principals
```diff
@@ -0,0 +1,42 @@
...
+local $::SERVICE_NAME = basename($0);
+local $::PROTOCOL_VERSION = "3.0.0";
+my $SCRIPT_VERSION = "3.0.0";
+
+perunServicesInit::init;
+my $data = perunServicesInit::getHashedHierarchicalData;
+my $DIRECTORY = perunServicesInit::getDirectory;
+
+
+our $A_PRINCIPALS;                    *A_PRINCIPALS =                       \'urn:perun:user:attribute-def:virt:kerberosLogins';
```
zlamalp, line 18: Service required attribute should be `user:def:kerberosLogins`

slave/process-kerberos-renewal-principals/changelog
```diff
@@ -0,0 +1 @@
+
```
zlamalp, line 1: Changelog must be filled with first version for the package to be built, so something like this should work:
```
perun-slave-process-kerberos-renewal-principals (3.0.0) stable; urgency=low

  * added new service kerberos-renewal-principals

 -- Johana Supikova <xsupikov@fi.muni.cz>  Wed, 11 Aug 2021 10:37:00 +0200
```

zlamalp (changes requested):

gen/kerberos_renewal_principals
Johaney-s: Switched `virt` to `def`.

zlamalp (changes requested):
Can you please rebase it? It will require new version in changelog for meta package.

Johaney-s:
> Can you please rebase it? It will require new version in changelog for meta package.

Rebased and updated changelog version.

zlamalp (approved):

PR #623, opened 2021-08-23
fix(slave): prevent unnecessary update of modification timestamps on k5login files

zlamalp:
If there is no change to the .k5login file, we shouldn't replace it.


zlamalp:
Tested and confirmed that it works as intended.

Johaney-s (approved):

PR #634, opened 2021-09-08
refactor(drupal_elixir): use only warning when duplicate users are found

zlamalp:
We need to end with OK return code in order to end with warning on duplicate users.


zlamalp:
Replaced with #754, closing.

PR #636, opened 2021-09-10
feat(vsup): Modified ad_user_vsup* scripts

Johaney-s:
* adjusted userPrincipalName, added proxyAddresses and exchangeMail

### DIFF
--- a/gen/ad_user_vsup_service
+++ b/gen/ad_user_vsup_service
@@ -7,7 +7,7 @@ use perunServicesUtils;
 
 local $::SERVICE_NAME = "ad_user_vsup_service";
 local $::PROTOCOL_VERSION = "3.0.1";
-my $SCRIPT_VERSION = "3.0.5";
+my $SCRIPT_VERSION = "3.0.6";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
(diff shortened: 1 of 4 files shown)


gen/ad_user_vsup_service
```diff
@@ -197,6 +200,10 @@ for my $login (@logins) {
 		print FILE "altSecurityIdentities: " . $val . "\n";
 	}
 
+	foreach my $val (@$$exchangeMailAliases) {
```
zlamalp, line 203: `@$$` should be `@$`

zlamalp (changes requested):

gen/ad_user_vsup_service
Johaney-s: Thanks, fixed.

zlamalp (approved):

send/ad_user_vsup
```diff
@@ -181,7 +181,7 @@ sub process_update() {
 			my $ad_entry = $ad_entries_map{$login};
 
 			# attrs without cn since it's part of DN to be updated
-			my @attrs = ('displayName','gecos','sn','givenName','mail','vsupPersonPersonalId','vsupPersonTitleHead','vsupPersonTitleTail','telephoneNumber','vsupPersonIdCardBarcode','vsupPersonIdCardChipNumber','altSecurityIdentities','accountExpires');
+			my @attrs = ('displayName','gecos','sn','givenName','mail','vsupPersonPersonalId','vsupPersonTitleHead','vsupPersonTitleTail','telephoneNumber','vsupPersonIdCardBarcode','vsupPersonIdCardChipNumber','altSecurityIdentities','accountExpires', 'userPrincipalName', 'proxyAddresses');
```
zlamalp, line 184: I realized, that you need to add those new attributes to the list on line 69, where you load entries from AD. Otherwise they are not present in entry and comparison will result in always setting new value. This applies also to loading entry by login in  `get_entry_by_login()` method. Also it needs to be updated in both send scripts (normal + service users).

zlamalp (changes requested):

send/ad_user_vsup
Johaney-s: I see. Added to both lists in both scripts.

zlamalp (approved):

PR #643, opened 2021-09-24
fix(send): log to STDERR when updating group membership in AD

metodej:
- Group membership update in AD now logs which members were
added/removed to STDERR as well as to local file.

### DIFF
--- a/send/ADConnector.pm
+++ b/send/ADConnector.pm
@@ -577,6 +577,7 @@ sub add_members_to_entry($$$$) {
 				ldap_log($service_name, "Group members added: " . $ad_entry->dn() . " | \n" . join(",\n", @$_));
 			} else {
 				ldap_log($service_name, "Group members NOT added: " . $ad_entry->dn() . " | " . $response->error() . " | \n" . join(",\n", @$_));
+				print STDERR "Group members NOT added: " . $ad_entry->dn() . " | " . $response->error() . " | \n" . join(",\n", @$_);
 				$return_code = $FAIL;
 			}
 		}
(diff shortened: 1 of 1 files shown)


Johaney-s (approved):

send/ADConnector.pm
```diff
@@ -575,8 +575,10 @@ sub add_members_to_entry($$$$) {
 		if ($response) {
 			unless ($response->is_error()) {
 				ldap_log($service_name, "Group members added: " . $ad_entry->dn() . " | \n" . join(",\n", @$_));
+				print STDERR "Group members added: " . $ad_entry->dn() . " | \n" . join(",\n", @$_);
```
cuadradek, line 578: I think we should only log errors to STDERR.

send/ADConnector.pm
```diff
@@ -609,8 +611,10 @@ sub remove_members_from_entry($$$$) {
 		if ($response) {
 			unless ($response->is_error()) {
 				ldap_log($service_name, "Group members removed: " . $ad_entry->dn() . " | \n" . join(",\n", @$_));
+				print STDERR "Group members removed: " . $ad_entry->dn() . " | \n" . join(",\n", @$_);
```
cuadradek, line 614: Same here.

send/ADConnector.pm
metodej: Yes, that would make more sense, thanks.

send/ADConnector.pm
metodej: Also removed.

cuadradek (approved):

PR #650, opened 2021-11-03
feat: new script ldap_it4i

metodej:
- added gen and send scripts for new service ldap_it4i to manage
updates of einfra users on IT4I LDAP

### DIFF
--- a/gen/ldap_it4i
+++ b/gen/ldap_it4i
@@ -1,0 +1,124 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use perunServicesInit;
+use perunServicesUtils;
+
+our $SERVICE_NAME = "ldap_it4i";
+local $::PROTOCOL_VERSION = "3.0.0";
+my $SCRIPT_VERSION = "3.0.0";
+
+perunServicesInit::init;
+my $DIRECTORY = perunServicesInit::getDirectory;
+my $fileName = "$DIRECTORY/$::SERVICE_NAME".".ldif";
+my $baseDnFileName = "$DIRECTORY/baseDN";
+
+my $data = perunServicesInit::getHashedHierarchicalData;
+
+# constants
+our $A_F_BASE_DN;  *A_F_BASE_DN = \'urn:perun:facility:attribute-def:def:ldapBaseDN';
+our $A_USER_LOGIN_EINFRA; *A_USER_LOGIN_EINFRA = \'urn:perun:user:attribute-def:def:login-namespace:einfra';
+our $A_FIRST_NAME;  *A_FIRST_NAME = \'urn:perun:user:attribute-def:core:firstName';
+our $A_LAST_NAME;  *A_LAST_NAME = \'urn:perun:user:attribute-def:core:lastName';
+our $A_DISPLAY_NAME;  *A_DISPLAY_NAME = \'urn:perun:user:attribute-def:core:displayName';
+our $A_USER_PREFERRED_MAIL;  *A_USER_PREFERRED_MAIL = \'urn:perun:user:attribute-def:def:preferredMail';
+our $A_SSHKEYS;  *A_SSHKEYS = \'urn:perun:user:attribute-def:def:sshPublicKey';
+our $A_MEMBER_STATUS;  *A_MEMBER_STATUS =  \'urn:perun:member:attribute-def:core:status';
+
+our $STATUS_VALID;      *STATUS_VALID =        \'VALID';
+our $STATUS_EXPIRED;    *STATUS_EXPIRED =      \'EXPIRED';
+our $STATUS_DISABLED;   *STATUS_DISABLED =     \'DISABLED';
+
+# check facility attribute
+my $ldapBaseDN = $data->getFacilityAttributeValue( attrName => $A_F_BASE_DN );
+if (!defined($ldapBaseDN)) {
+	exit 1;
+}
+
+# gather user data
+my $users;
+
+foreach my $resourceId ( $data->getResourceIds() ) {
+	foreach my $memberId ( $data->getMemberIdsForResource( resource => $resourceId )) {
+
+		my $login = $data->getUserAttributeValue( member => $memberId, attrName => $A_USER_LOGIN_EINFRA );
+		$users->{$login}->{$A_FIRST_NAME} = $data->getUserAttributeValue( member => $memberId, attrName => $A_FIRST_NAME );
+		$users->{$login}->{$A_LAST_NAME} = $data->getUserAttributeValue( member => $memberId, attrName => $A_LAST_NAME );
+		$users->{$login}->{$A_DISPLAY_NAME} = $data->getUserAttributeValue( member => $memberId, attrName => $A_DISPLAY_NAME );
+		$users->{$login}->{$A_USER_PREFERRED_MAIL} = $data->getUserAttributeValue( member => $memberId, attrName => $A_USER_PREFERRED_MAIL );
+		$users->{$login}->{$A_SSHKEYS} = $data->getUserAttributeValue( member => $memberId, attrName => $A_SSHKEYS );
+
+		my $status = $data->getMemberAttributeValue( member => $memberId, attrName => $A_MEMBER_STATUS );
+		if ($status eq $STATUS_VALID) {
+			$status = 'ACTIVE';
+		} elsif ($status eq $STATUS_EXPIRED) {
+			$status = 'INACTIVE';
+		} elsif ($status eq $STATUS_DISABLED) {
+			$status = 'ARCHIVED';
+		}
+		$users->{$login}->{$A_MEMBER_STATUS} = $status;
+	}
+}
+
+# print BASE_DN file
+open FILE,">:encoding(UTF-8)","$baseDnFileName" or die "Cannot open $baseDnFileName: $! \n";
+print FILE $ldapBaseDN;
+close(FILE);
+
+# print user data LDIF
+open FILE,">:encoding(UTF-8)","$fileName" or die "Cannot open $fileName: $! \n";
+
+for my $login (sort keys %$users) {
+
+	my $givenName = $users->{$login}->{$A_FIRST_NAME};
+	my $sn = $users->{$login}->{$A_LAST_NAME};
+
+	print FILE "dn: uid=" . $login . "," . $ldapBaseDN . "\n";
+	print FILE "uid: " . $login . "\n";
+
+	if (defined $givenName and length $givenName) {
+		print FILE "givenName: " . $givenName . "\n";
+	}
+
+	if (defined $sn and length $sn) {
+		print FILE "sn: " . $sn . "\n";
+	}
+
+	if (defined $sn and length $sn and defined $givenName and length $givenName) {
+		print FILE "cn: " . $givenName . " " . $sn . "\n";
+	} elsif (defined $sn and length $sn) {
+		print FILE "cn: " . $sn . "\n";
+	} elsif (defined $givenName and length $givenName) {
+		print FILE "cn: " . $givenName . "\n";
+	} else {
+		print FILE "cn: N/A\n";
+	}
+
+	print FILE "displayName: " . $users->{$login}->{$A_DISPLAY_NAME} . "\n";
+	print FILE "mail: " . $users->{$login}->{$A_USER_PREFERRED_MAIL} . "\n";
+	print FILE "userPassword: {SASL}" . $login . '@EINFRA' . "\n";
+	print FILE "status: " . $users->{$login}->{$A_MEMBER_STATUS} . "\n";
+
+	my $sshKeys = $users->{$login}->{$A_SSHKEYS};
+	if (defined $sshKeys and length $sshKeys) {
+		foreach my $sshKey (sort @$sshKeys) {
+			print FILE "sshPublicKey: " . $sshKey . "\n";
+		}
+	}
+
+	# print classes
+	print FILE "objectclass: top\n";
+	print FILE "objectclass: person\n";
+	print FILE "objectclass: einfraPerson\n";
+	print FILE "objectclass: inetOrgPerson\n";
+	print FILE "objectclass: ldapPublicKey\n";
+
+	# there must be empty line after each entry
+	print FILE "\n";
+
+}
+
+close FILE;
+
+perunServicesInit::finalize;
--- a/send/ldap_it4i
+++ b/send/ldap_it4i
@@ -1,0 +1,165 @@
+#!/usr/bin/perl
+use strict;
+use warnings;
+no if $] >= 5.017011, warnings => 'experimental::smartmatch';
+use Net::LDAPS;
+use Net::LDAP::Entry;
+use Net::LDAP::Message;
+use Net::LDAP::LDIF;
+use ScriptLock;
+use ADConnector;
+
+sub process_update;
+sub process_archive;
+
+# Constants and counters
+my $counter_add = 0;
+my $counter_update = 0;
+my $counter_archive = 0;
+my $counter_fail = 0;
+
+# Define service
+my $service_name = "ldap_it4i";
+
+# GEN folder location
+my $facility_name = $ARGV[0];
+chomp($facility_name);
+my $service_files_base_dir="../gen/spool";
+my $service_files_dir="$service_files_base_dir/$facility_name/$service_name";
+my $service_file = "$service_files_dir/$service_name.ldif";
+
+# BASE DN
+open my $file, '<', "$service_files_dir/baseDN";
+my $base_dn = <$file>;
+chomp($base_dn);
+close $file;
+
+# propagation destination
+my $namespace = $ARGV[1];
+chomp($namespace);
+
+# create service lock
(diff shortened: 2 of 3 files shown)


send/ldap_it4i
```diff
@@ -0,0 +1,164 @@
...
+					);
+				}
+			}
+
+			if (@entry_changed) {
+				# Update entry in AD
+				my $response = $previousUser->update($ldap);
+				unless ($response->is_error()) {
+					# SUCCESS
+					ldap_log($service_name, "Added: " . $previousUser->dn());
```
cuadradek, line 121:
```suggestion
					ldap_log($service_name, "Updated: " . $previousUser->dn());
```

cuadradek (changes requested):

gen/ldap_it4i
```diff
@@ -0,0 +1,99 @@
...
+# print user data LDIF
+open FILE,">:encoding(UTF-8)","$fileName" or die "Cannot open $fileName: $! \n";
+
+for my $login (sort keys %$users) {
+
+	print FILE "dn: uid=" . $login . "," . $ldapBaseDN . "\n";
+	print FILE "uid: " . $login . "\n";
+	print FILE "givenName: " . $users->{$login}->{$A_FIRST_NAME} . "\n";
+	print FILE "sn: " . $users->{$login}->{$A_LAST_NAME} . "\n";
+	print FILE "fullName: " . $users->{$login}->{$A_FIRST_NAME} . " " . $users->{$login}->{$A_LAST_NAME} . "\n";
```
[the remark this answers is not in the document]
metodej: I've missed that, thanks!

send/ldap_it4i
```diff
@@ -0,0 +1,164 @@
...
+my $ldap_location = resolve_pdc($conf[0]);
+my $ldap = ldap_connect($ldap_location);
+my $filter = '(objectClass=person)';
+
+# connect
+ldap_bind($ldap, $conf[1], $conf[2]);
+
+# load all data
+my @users = load_perun($service_file);
+my @previousUsers = load_ad($ldap, $base_dn, $filter, ['uid','givenName','sn','displayName','mail','userPassword','state','sshPublicKey']);
```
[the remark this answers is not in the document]
metodej: Fixed.

send/ldap_it4i
metodej: Fixed.

send/ldap_it4i
```diff
@@ -0,0 +1,164 @@
...
+			if (@entry_changed) {
+				# Update entry in AD
+				my $response = $previousUser->update($ldap);
+				unless ($response->is_error()) {
+					# SUCCESS
+					ldap_log($service_name, "Added: " . $previousUser->dn());
+					$counter_update++;
+				} else {
+					# FAIL
+					ldap_log($service_name, "NOT added: " . $previousUser->dn() . " | " . $response->error());
```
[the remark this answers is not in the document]
metodej: Fixed.

Johaney-s (approved):

cuadradek (approved):

gen/ldap_it4i
```diff
@@ -0,0 +1,100 @@
...
+		$users->{$login}->{$A_FIRST_NAME} = $data->getUserAttributeValue( member => $memberId, attrName => $A_FIRST_NAME );
+		$users->{$login}->{$A_LAST_NAME} = $data->getUserAttributeValue( member => $memberId, attrName => $A_LAST_NAME );
+		$users->{$login}->{$A_DISPLAY_NAME} = $data->getUserAttributeValue( member => $memberId, attrName => $A_DISPLAY_NAME );
+		$users->{$login}->{$A_USER_PREFERRED_MAIL} = $data->getUserAttributeValue( member => $memberId, attrName => $A_USER_PREFERRED_MAIL );
+
+		my $status = $data->getMemberAttributeValue( member => $memberId, attrName => $A_MEMBER_STATUS );
+		if ($status eq $STATUS_VALID) {
+			$status = 'ACTIVE';
+		} elsif ($status eq $STATUS_INVALID) {
+			$status = 'INACTIVE';
```
cuadradek, line 55: I think old states are mapped to new states like this:
VALID -> ACTIVE
INVALID -> INCOMPLETE
EXPIRED -> INACTIVE
DISABLED -> ARCHIVED

cuadradek (changes requested):

gen/ldap_it4i
metodej: Fixed, thanks.

cuadradek (approved):

gen/ldap_it4i
```diff
@@ -0,0 +1,103 @@
...
+	print FILE "sn: " . $users->{$login}->{$A_LAST_NAME} . "\n";
+	print FILE "fullName: " . $users->{$login}->{$A_FIRST_NAME} . " " . $users->{$login}->{$A_LAST_NAME} . "\n";
+	print FILE "displayName: " . $users->{$login}->{$A_DISPLAY_NAME} . "\n";
+	print FILE "mail: " . $users->{$login}->{$A_USER_PREFERRED_MAIL} . "\n";
+	print FILE "userPassword: {SASL}" . $login . '@EINFRA' . "\n";
+	print FILE "state: " . $users->{$login}->{$A_MEMBER_STATUS} . "\n";
+
+	my $sshKey = $users->{$login}->{$A_SSHKEY};
+	if (defined $sshKey and length $sshKey) {
+		print FILE "sshPublicKey: " . $sshKey . "\n";
```
zlamalp, line 93: SSH key attribute contains array. You must print it in a cycle on multiple lines. See other LDAP/AD services, where multi-value attributes are printed.
Also I think you can simplify original variable assignment when constructing `$users` structure. There is no necessity to check if its defined or non-empty, you can do it when printing the value here.

zlamalp (changes requested):

send/ldap_it4i
```diff
@@ -0,0 +1,164 @@
+#!/usr/bin/perl
+use strict;
+use warnings;
+no if $] >= 5.017011, warnings => 'experimental::smartmatch';
+use Net::LDAPS;
+use Net::LDAP::Entry;
+use Net::LDAP::Message;
+use Net::LDAP::LDIF;
+use ScriptLock;
```
zlamalp, line 9: You should add `use ADConnector;` if you are using this library methods.

send/ldap_it4i
```diff
@@ -0,0 +1,164 @@
...
+my $namespace = $ARGV[1];
+chomp($namespace);
+
+# create service lock
+my $lock = ScriptLock->new($facility_name . "_" . $service_name . "_" . $namespace);
+($lock->lock() == 1) or die "Unable to get lock, service propagation was already running.";
+
+# init configuration
+my @conf = init_config($namespace);
+my $ldap_location = resolve_pdc($conf[0]);
```
zlamalp, line 46: We can't use function to resolve PDC (primary domain controller) from DNS, since the LDAP instance is not an AD. We should use raw value from the config (`$conf[0]`). Please make note about it in the script itself.

zlamalp (changes requested):

gen/ldap_it4i
metodej: Ok, thanks, I've also added sorting to compare in send script...

gen/ldap_it4i
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	# there must be empty line after each entry
+	print FILE "\n";
```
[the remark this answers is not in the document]
metodej: Added.

send/ldap_it4i
metodej: Added.

send/ldap_it4i
metodej: Added.

gen/ldap_it4i
```diff
@@ -0,0 +1,105 @@
...
+	print FILE "givenName: " . $users->{$login}->{$A_FIRST_NAME} . "\n";
+	print FILE "sn: " . $users->{$login}->{$A_LAST_NAME} . "\n";
+	print FILE "fullName: " . $users->{$login}->{$A_FIRST_NAME} . " " . $users->{$login}->{$A_LAST_NAME} . "\n";
+	print FILE "displayName: " . $users->{$login}->{$A_DISPLAY_NAME} . "\n";
+	print FILE "mail: " . $users->{$login}->{$A_USER_PREFERRED_MAIL} . "\n";
+	print FILE "userPassword: {SASL}" . $login . '@EINFRA' . "\n";
+	print FILE "state: " . $users->{$login}->{$A_MEMBER_STATUS} . "\n";
+
+	my $sshKeys = $users->{$login}->{$A_SSHKEYS};
+	foreach my $sshKey (@$sshKeys) {
```
zlamalp, line 88: I think simple retyping to array will fail if `$sshKeys` is `undef`. Maybe add check for it (wrap cycle in IF). Also you can sort ssh keys themselves, so they don't differ between two script runs.

send/ldap_it4i
```diff
@@ -0,0 +1,166 @@
...
+# private methods
+sub process_update() {
+
+	foreach my $user (@users) {
+
+		if (exists $previousUsersMap{$user->get_value('uid')}) {
+
+			my $previousUser = $previousUsersMap{$user->get_value('uid')};
+
+			# attrs without cn since it's part of DN to be updated
```
zlamalp, line 97: We are using `uid` as primary attribute used in DN, so you can add `cn` between attributes for update and fix this comment.

Please update `load_perun()` and `load_ad()` functions in `ADConnector.pm`. There are checks for required attributes of entry, without which entries are discarded from returned result. We must add `uid` between conditions in order to get entries for this service to be processed. Thank you.

zlamalp (changes requested):

gen/ldap_it4i
```diff
@@ -0,0 +1,105 @@
...
+		$users->{$login}->{$A_FIRST_NAME} = $data->getUserAttributeValue( member => $memberId, attrName => $A_FIRST_NAME );
+		$users->{$login}->{$A_LAST_NAME} = $data->getUserAttributeValue( member => $memberId, attrName => $A_LAST_NAME );
+		$users->{$login}->{$A_DISPLAY_NAME} = $data->getUserAttributeValue( member => $memberId, attrName => $A_DISPLAY_NAME );
+		$users->{$login}->{$A_USER_PREFERRED_MAIL} = $data->getUserAttributeValue( member => $memberId, attrName => $A_USER_PREFERRED_MAIL );
+		$users->{$login}->{$A_SSHKEYS} = $data->getUserAttributeValue( member => $memberId, attrName => $A_SSHKEYS );
+
+		my $status = $data->getMemberAttributeValue( member => $memberId, attrName => $A_MEMBER_STATUS );
+		if ($status eq $STATUS_VALID) {
+			$status = 'ACTIVE';
+		} elsif ($status eq $STATUS_INVALID) {
```
zlamalp, line 56: INVALID members shouldn't be part of the export, we don't have to process this status here.

gen/ldap_it4i
```diff
@@ -0,0 +1,105 @@
...
+
+	print FILE "dn: uid=" . $login . "," . $ldapBaseDN . "\n";
+	print FILE "uid: " . $login . "\n";
+	print FILE "givenName: " . $users->{$login}->{$A_FIRST_NAME} . "\n";
+	print FILE "sn: " . $users->{$login}->{$A_LAST_NAME} . "\n";
+	print FILE "fullName: " . $users->{$login}->{$A_FIRST_NAME} . " " . $users->{$login}->{$A_LAST_NAME} . "\n";
+	print FILE "displayName: " . $users->{$login}->{$A_DISPLAY_NAME} . "\n";
+	print FILE "mail: " . $users->{$login}->{$A_USER_PREFERRED_MAIL} . "\n";
+	print FILE "userPassword: {SASL}" . $login . '@EINFRA' . "\n";
+	print FILE "state: " . $users->{$login}->{$A_MEMBER_STATUS} . "\n";
```
zlamalp, line 85: Attribute `state` will be most probably renamed to `status`. I'll let you know once it will be decided.

zlamalp (changes requested):

gen/ldap_it4i
zlamalp: Change is confirmed, it will be `status`, please update attribute name in both gen and send scripts.

gen/ldap_it4i
```diff
@@ -0,0 +1,105 @@
...
+# print user data LDIF
+open FILE,">:encoding(UTF-8)","$fileName" or die "Cannot open $fileName: $! \n";
+
+for my $login (sort keys %$users) {
+
+	print FILE "dn: uid=" . $login . "," . $ldapBaseDN . "\n";
+	print FILE "uid: " . $login . "\n";
+	print FILE "givenName: " . $users->{$login}->{$A_FIRST_NAME} . "\n";
+	print FILE "sn: " . $users->{$login}->{$A_LAST_NAME} . "\n";
+	print FILE "fullName: " . $users->{$login}->{$A_FIRST_NAME} . " " . $users->{$login}->{$A_LAST_NAME} . "\n";
```
zlamalp, line 81: Instead of `fullName` this will be `cn`. But attribute is required by schema, so the script should handle empty names more gracefully. See other services for AD, where value is constructed differently (usually for the displayName) where we check for empty values and if result is empty, its replaced with `N/A` (this is probably only in LDAPc).

send/ldap_it4i
```diff
@@ -0,0 +1,166 @@
...
+my $ldap_location = $conf[0];
+my $ldap = ldap_connect($ldap_location);
+my $filter = '(objectClass=person)';
+
+# connect
+ldap_bind($ldap, $conf[1], $conf[2]);
+
+# load all data
+my @users = load_perun($service_file);
+my @previousUsers = load_ad($ldap, $base_dn, $filter, ['uid','givenName','sn','fullName','displayName','mail','userPassword','state','sshPublicKey']);
```
zlamalp, line 57: Based on other comments, there should be `fullName` replaced with `cn` and `state` with `status`. Also we probably do not need to get/update `userPassword`, since users shouldn't have their login changed! It would be considered as different entry by the script anyway.

send/ldap_it4i
```diff
@@ -0,0 +1,166 @@
...
+sub process_update() {
+
+	foreach my $user (@users) {
+
+		if (exists $previousUsersMap{$user->get_value('uid')}) {
+
+			my $previousUser = $previousUsersMap{$user->get_value('uid')};
+
+			# attrs without cn since it's part of DN to be updated
+			my @attrs = ('givenName','sn','fullName','displayName','mail','userPassword','state','sshPublicKey');
```
zlamalp, line 98: Again, replace `fullName` with `cn`, `state` with `status` and remove `userPassword`. Thank you.

gen/ldap_it4i
```diff
@@ -0,0 +1,105 @@
...
+	my $sshKeys = $users->{$login}->{$A_SSHKEYS};
+	foreach my $sshKey (@$sshKeys) {
+		print FILE "sshPublicKey: " . $sshKey . "\n";
+	}
+
+	# print classes
+	print FILE "objectclass: top\n";
+	print FILE "objectclass: person\n";
+	print FILE "objectclass: inetOrgPerson\n";
+	print FILE "objectclass: ldapPublicKey\n";
```
zlamalp, line 96: Please add `einfraPerson` between `objectClass` values. Thank you.

zlamalp (changes requested):

gen/ldap_it4i
metodej: I have tested this and even for `undef` value, the cycle just does not go through any values and does not fail, so I do not think the check is necessary. Also there is a function in ADConnector for comparing unsorted values used in most of the ad scripts, I have just switch to this instead of plain `~~`, I think it is better than sorting the values in gen script. Or do you think I should sort them anyway?

send/ldap_it4i
metodej: Ok, added, thanks.

gen/ldap_it4i
metodej: Removed.

gen/ldap_it4i
metodej: Changed.

gen/ldap_it4i
metodej: Added check of `emptiness` of `givenName` and sn values and the cn value is `N/A` if both of them are empty.

send/ldap_it4i
metodej: Fixed.

send/ldap_it4i
metodej: Fixed.

gen/ldap_it4i
metodej: Fixed.

gen/ldap_it4i
zlamalp: Then it should be OK, regarding the `undef`. Sorting in gen script is necessary when we want to compare gen outputs (eg. when changing the gen script).

metodej: Good point, I've added the sort and in that case, the check is needed.

zlamalp (approved):

PR #657, opened 2021-12-14
feat(vmware_ldap): generic service to manage LDAP for VMware

zlamalp:
- This service is a copy of ldap_vsb_vi with updated
  service name, protocol/scrip versions and variable namings.
- It will be used as generic service for multiple
  VMware instances and will replace the original service
  in the future.

### DIFF
--- a/gen/vmware_ldap
+++ b/gen/vmware_ldap
@@ -1,0 +1,203 @@
+#!/usr/bin/perl
+use strict;
+use warnings;
+use perunServicesInit;
+use perunServicesUtils;
+use Text::Unidecode;
+
+local $::SERVICE_NAME = "vmware_ldap";
+local $::PROTOCOL_VERSION = "1.0.0";
+my $SCRIPT_VERSION = "1.0.0";
+
+perunServicesInit::init;
+my $DIRECTORY = perunServicesInit::getDirectory;
+my $fileName_users = "$DIRECTORY/$::SERVICE_NAME"."_users".".ldif";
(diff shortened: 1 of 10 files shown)


gen/vmware_ldap
```diff
@@ -0,0 +1,203 @@
+#!/usr/bin/perl
+use strict;
+use warnings;
+use perunServicesInit;
+use perunServicesUtils;
+use Text::Unidecode;
+
+local $::SERVICE_NAME = "vmware_ldap";
+local $::PROTOCOL_VERSION = "1.0.0";
+my $SCRIPT_VERSION = "1.0.0";
```
metodej, line 10: Shouldn't these versions be `"3.0.0"`?

slave/process-vmware-ldap/bin/process-vmware_ldap.sh
```diff
@@ -0,0 +1,103 @@
+#!/bin/bash
+
+# Original update script created by (C) funny from ZCU
+# 2007-08-05 handle bug in ldapsearch
+# 2011-08-10 generification of script
+# 2013-06-20 prevent concurrent run
+
+PROTOCOL_VERSION='1.0.0'
```
metodej, line 8: Same here.

gen/vmware_ldap
metodej: Okay, makes sense.

metodej (approved):

PR #668, opened 2022-03-07
feat(lawdb_export_mu): add new lawdb_export_mu service

Johaney-s:
* added gen and send scripts


Johaney-s (commented):
As I commited this on Michal's behalf Git does not allow me to approve this, but besides probably redundant group id attribute constant in gen script I see no problem.

HejdaJakub (approved):

PR #676, opened 2022-04-05
feat: added local changes and unversioned scripts from idm instance

xkostka2:
* added gen/send script for elerning_mu_ff, which fills data to elerning portal of FF
* added gen/send script for teams_provisioning, which manages provisioning to MS Teams
* changed limit on entry updates in send script crm_ceitec
* added local changes from idm instance to send script ad_user_mu_ucn
* changed path to auth token in send script for netbox

### DIFF
--- a/gen/elerning_mu_ff
+++ b/gen/elerning_mu_ff
@@ -1,0 +1,63 @@
+#!/usr/bin/perl
+use strict;
+use warnings;
+use perunServicesInit;
+use perunServicesUtils;
+use File::Basename;
+
+our $SERVICE_NAME = basename($0);
+our $PROTOCOL_VERSION = "3.0.1";
+my $SCRIPT_VERSION = "3.0.2";
+
+perunServicesInit::init;
+my $DIRECTORY = perunServicesInit::getDirectory;
+my $data = perunServicesInit::getHashedHierarchicalData;
+
+#Constants
+our $A_USER_LOGIN;          		 *A_USER_LOGIN =            		 \'urn:perun:user:attribute-def:def:login-namespace:mu';
+our $A_USER_FIRST_NAME;              *A_USER_FIRST_NAME =                \'urn:perun:user:attribute-def:core:firstName';
+our $A_USER_LAST_NAME;               *A_USER_LAST_NAME =                 \'urn:perun:user:attribute-def:core:lastName';
+our $A_USER_STATUS;                  *A_USER_STATUS =                    \'urn:perun:member:attribute-def:core:status';
+
+our $STATUS_VALID;                   *STATUS_VALID =                     \'VALID';
+
+
+my $fileName = "$DIRECTORY/$::SERVICE_NAME";
+
+my $users;
+#####################################
+
+foreach my $resourceId ($data->getResourceIds()) {
+	foreach my $memberId ($data->getMemberIdsForResource(resource => $resourceId)) {
+		my $memberStatus = $data->getMemberAttributeValue(member=> $memberId, attrName=>$A_USER_STATUS);
+		my $memberLogin = $data->getUserAttributeValue(member=> $memberId, attrName=>$A_USER_LOGIN);
+		my $memberAttributes->{$A_USER_FIRST_NAME} =  $data->getUserAttributeValue(member=> $memberId, attrName=>$A_USER_FIRST_NAME);
+		$memberAttributes->{$A_USER_LAST_NAME} =  $data->getUserAttributeValue(member=> $memberId, attrName=>$A_USER_LAST_NAME);
+
+		if($memberStatus eq $STATUS_VALID) {
+			$users->{$memberLogin} = $memberAttributes;
+		}
+
+	}
+}
+
+####### output ######################
+open FILE,">$fileName" or die "Cannot open $fileName: $! \n";
+binmode FILE, ":utf8";
+
+for my $login (keys %$users) {
+
+	my $firstName = $users->{$login}->{$A_USER_FIRST_NAME};
+	$firstName =~ s/'/''/g;
+	my $lastName = $users->{$login}->{$A_USER_LAST_NAME};
+	$lastName =~ s/'/''/g;
+
+	print FILE 
+			qq/UPDATE [dbo].[tbl_User] SET [username]='$login', [FirstName]='$firstName', [LastName]='$lastName', [EmailAddress]='$login\@mail.muni.cz'  WHERE username='$login'/, "\n",
+			qq/IF \@\@ROWCOUNT = 0/, "\n",
+			qq/INSERT INTO [dbo].[tbl_User] ([Username],[Password],[FirstName],[LastName],[EmailAddress],[System],[MediaFormat],[LoginCount],[ProfilePicFilePath],[ProfilePicExtension],[ProfilePicReported],[PlaylistEnabled],[DefaultQuality],[Language],[status]) VALUES/, "\n",
+			qq/('$login', '', '$firstName', '$lastName', '$login\@mail.muni.cz', 0, 'F', 0, 'default', '.jpg', 0, 0, 'hi', 'en', 1);/, "\n\n";
+}
+close FILE or die "Cannot close $fileName: $! \n";
+#####################################################
+perunServicesInit::finalize;
--- a/gen/teams_provisioning
+++ b/gen/teams_provisioning
@@ -1,0 +1,44 @@
+#!/usr/bin/perl
+use strict;
+use warnings;
+use perunServicesInit;
+use perunServicesUtils;
+use Perun::Agent;
+use JSON::XS;
+use Data::Dumper;
+
+our $SERVICE_NAME = "teams_provisioning";
+our $PROTOCOL_VERSION = "3.0.0";
+my $SCRIPT_VERSION = "3.0.1";
+
+perunServicesInit::init;
+my $DIRECTORY = perunServicesInit::getDirectory;
+my $data = perunServicesInit::getHashedHierarchicalData;
+
+#Constants
+our $A_RES_ALIAS;             *A_RES_ALIAS =         \'urn:perun:resource:attribute-def:def:o365ResourceAlias';
+our $A_USER_LOGIN;            *A_USER_LOGIN=             \'urn:perun:user:attribute-def:def:login-namespace:mu';
+
+my @teams = ();
+foreach my $resourceId ($data->getResourceIds()) {
+		my $resAlias = $data->getResourceAttributeValue(resource=>$resourceId, attrName=>$A_RES_ALIAS);
+        print $resAlias . "\n";
+        my $team->{"identity"} = $resAlias;
+        my @members = ();
+        foreach my $memberId ($data->getMemberIdsForResource(resource => $resourceId)) {
+			my $userLogin = $data->getUserAttributeValue(member=> $memberId, attrName=>$A_USER_LOGIN);
+            print $userLogin . "\n";
+            push @members, $userLogin;
+        };
+		@members = sort(@members);
+        $team->{"members"} = \@members;
+        $team->{"owners"} = ();
+        push @teams, $team;
+};
+
+my $fileName = "$DIRECTORY/teams.json";
+open FILE, ">$fileName" or die "Cannot open $fileName: $! \n";
+print FILE JSON::XS->new->utf8->pretty->canonical->encode(\@teams);
+close FILE or die "Cannot close $fileName: $! \n";
+
+perunServicesInit::finalize;
--- a/send/ad_user_mu_ucn
+++ b/send/ad_user_mu_ucn
@@ -14,12 +14,16 @@ use ScriptLock;
 
 sub process_add;
 sub process_update;
+sub process_remove;
 sub ping_password_setter;
 
 # log counters
 my $counter_add = 0;
+my $counter_add_fail = 0;
 my $counter_update = 0;
-my $counter_fail = 0;
+my $counter_update_fail = 0;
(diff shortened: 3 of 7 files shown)


send/ad_user_mu_ucn
```diff
@@ -191,14 +234,15 @@ sub process_update() {
 				unless ($response->is_error()) {
 					# SUCCESS
 					foreach my $log_message (@entry_changed) {
-						ldap_log($service_name, "Updated: " . $ad_entry->dn() . " | " . $log_message);
+						#ldap_log($service_name, "Updated: " . $ad_entry->dn() . " | " . $log_message);
```
stavamichal, line 237: There should be a comment about this behavior. Because there are too many update operations, we don't want to spam the log if not necessary.

send/arcgis_mu
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 SERVICE_NAME="arcgis_mu"
 
-. generic_send
+. generic_send2
```
stavamichal, line 4: This is wrong, you can remove it from here. Fixed in the original file.

send/o365_mu_process.pl
```diff
@@ -11,22 +11,16 @@
 use File::Copy;
 use File::Path qw(make_path);
 use Data::Dumper;
-use JSON qw( decode_json );
```
stavamichal, line 14: This whole script should be skipped because this is a local change.

send/openstack_projects
```diff
@@ -40,4 +40,3 @@ else
 fi
 
 exit $ERR_CODE
-
```
stavamichal, line 43: This is a not needed change. It can be replaced by the original script with the right end of the file.

send/crm_ceitec.php
```diff
@@ -18,7 +18,7 @@
 
 $params = getopt("", $longopts);
 
-# public string[] CreateUser ( string userName, string firstName, string lastName, string email, string orgUnit, string universityId, string eppn , string rgs, string alternativeLoginNames )
+# public string[] CreateUser ( string userName, string firstName, string lastName, string email, string orgUnit, string universityId, string eppn , string rgs , string alternativeLoginNames )
```
stavamichal, line 21: Not needed change, you can use the original file instead.

stavamichal (changes requested):

gen/ad_admin_user_mu_ucn
```diff
@@ -115,4 +115,13 @@ for my $login (@logins) {
 
 close(FILE);
 
+sub checkBase64 {
```
zlamalp, line 118: This added method is not actually used by the script, so I'm not sure we should add it.

gen/elerning_mu_ff
```diff
@@ -0,0 +1,64 @@
...
+use perunServicesUtils;
+use File::Basename;
+
+our $SERVICE_NAME = basename($0);
+our $PROTOCOL_VERSION = "3.0.1";
+my $SCRIPT_VERSION = "3.0.1";
+
+perunServicesInit::init;
+my $DIRECTORY = perunServicesInit::getDirectory;
+my $data = perunServicesInit::getHierarchicalData;
```
zlamalp, line 14: Script should be converted to usage of `getHashedHierarchicalData()`.

gen/teams_provisioning
```diff
@@ -0,0 +1,45 @@
...
+use JSON::XS;
+use Data::Dumper;
+
+our $SERVICE_NAME = "teams_provisioning";
+our $PROTOCOL_VERSION = "3.0.0";
+my $SCRIPT_VERSION = "3.0.0";
+
+perunServicesInit::init;
+my $DIRECTORY = perunServicesInit::getDirectory;
+my $data = perunServicesInit::getHierarchicalData;
```
zlamalp, line 16: Script should be converted to `getHashedHierarchicalData()`.

send/elerning_mu_ff
```diff
@@ -0,0 +1,41 @@
...
+		DESTINATION_DIR="$DESTINATION"
+		;;
+	*)
+		echo "Unknown destination type '$DESTINATION_TYPE'." >&2
+		exit 1;
+		;;
+esac
+
+
+	TIMESTAMP=`date "+%F_%H.%M.%S"`
```
zlamalp, line 39: I think indents of the two last commands are unnecessary.

zlamalp (changes requested):

gen/elerning_mu_ff
```diff
@@ -0,0 +1,64 @@
...
+our $A_USER_STATUS;                  *A_USER_STATUS =                    \'urn:perun:member:attribute-def:core:status';
+
+our $STATUS_VALID;                   *STATUS_VALID =                     \'VALID';
+
+
+my $fileName = "$DIRECTORY/$::SERVICE_NAME";
+
+my $users;
+
+my %facilityAttributes = attributesToHash $data->getAttributes;
```
zlamalp, line 29: Whole processing of `$data` has to be updated to work with the new hashed data, since the returned JSON structure is different. See reworked scripts and do the similar like in #590.

Also please update script version afterwards, eg. `3.0.1` -> `3.0.2`.

gen/teams_provisioning
```diff
@@ -0,0 +1,45 @@
...
+perunServicesInit::init;
+my $DIRECTORY = perunServicesInit::getDirectory;
+my $data = perunServicesInit::getHashedHierarchicalData;
+
+#Constants
+our $A_RES_ALIAS;             *A_RES_ALIAS =         \'urn:perun:resource:attribute-def:def:o365ResourceAlias';
+our $A_USER_LOGIN;            *A_USER_LOGIN=             \'urn:perun:user:attribute-def:def:login-namespace:mu';
+
+my @teams = ();
+my @resourcesData = $data->getChildElements;
```
zlamalp, line 23: Whole processing of `$data` has to be updated to work with the new hashed data, since the returned JSON structure is different. See reworked scripts and do the similar like in #590.

Also please update script version afterwards, eg. `3.0.1` -> `3.0.2`.

gen/elerning_mu_ff
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/perl
```
zlamalp, line 1: I think there is missing `a` in the service name, as it should be "elearning". Technically it works, since the service name is the same everywhere, but maybe discuss it with Michal if he wants to fix it.

zlamalp (changes requested):

gen/elerning_mu_ff
stavamichal: It should be e-learning but it is in so many places at this moment that I would let it for a further fix.

gen/elerning_mu_ff
```diff
@@ -0,0 +1,63 @@
...
+
+my $users;
+#####################################
+
+foreach my $resourceId ($data->getResourceIds()) {
+	foreach my $memberId ($data->getMemberIdsForResource($resourceId)) {
+		my $memberStatus = $data->getMemberAttributeValue(member=> $memberId, attrName=>$A_USER_STATUS);
+		my $memberLogin = $data->getMemberAttributeValue(member=> $memberId, attrName=>$A_USER_FACILITY_LOGIN);
+		my %memberAttributes->{$A_USER_FIRST_NAME} =  $data->getMemberAttributeValue(member=> $memberId, attrName=>$A_USER_FIRST_NAME);
+		%memberAttributes->{$A_USER_LAST_NAME} =  $data->getMemberAttributeValue(member=> $memberId, attrName=>$A_USER_LAST_NAME);
```
zlamalp, line 35: Those are user attributes, you have to use `getUserAttributeValue()`.

gen/elerning_mu_ff
```diff
@@ -0,0 +1,63 @@
...
+my $fileName = "$DIRECTORY/$::SERVICE_NAME";
+
+my $users;
+#####################################
+
+foreach my $resourceId ($data->getResourceIds()) {
+	foreach my $memberId ($data->getMemberIdsForResource($resourceId)) {
+		my $memberStatus = $data->getMemberAttributeValue(member=> $memberId, attrName=>$A_USER_STATUS);
+		my $memberLogin = $data->getMemberAttributeValue(member=> $memberId, attrName=>$A_USER_FACILITY_LOGIN);
+		my %memberAttributes->{$A_USER_FIRST_NAME} =  $data->getMemberAttributeValue(member=> $memberId, attrName=>$A_USER_FIRST_NAME);
```
zlamalp, line 34: You can use scalar context for this hash here and in all later usages, since you are addressing specific keys directly. `%memberAttributes` -> `$memberAttributes`.

gen/elerning_mu_ff
```diff
@@ -0,0 +1,63 @@
...
+		%memberAttributes->{$A_USER_LAST_NAME} =  $data->getMemberAttributeValue(member=> $memberId, attrName=>$A_USER_LAST_NAME);
+
+		if($memberStatus eq $STATUS_VALID) {
+			$users->{$memberLogin} = \%memberAttributes;
```
zlamalp, line 38: `\%memberAttributes` -> `$memberAttributes`

gen/teams_provisioning
```diff
@@ -0,0 +1,43 @@
...
+use warnings;
+use perunServicesInit;
+use perunServicesUtils;
+use Perun::Agent;
+use JSON::XS;
+use Data::Dumper;
+
+our $SERVICE_NAME = "teams_provisioning";
+our $PROTOCOL_VERSION = "3.0.2";
+my $SCRIPT_VERSION = "3.0.2";
```
zlamalp, line 12: Original version is `3.0.0` for this service so you should use `3.0.1` now. Also protocol version should stay at `3.0.0` since content of the generated files has not changed.

gen/teams_provisioning
```diff
@@ -0,0 +1,43 @@
...
+my $DIRECTORY = perunServicesInit::getDirectory;
+my $data = perunServicesInit::getHashedHierarchicalData;
+
+#Constants
+our $A_RES_ALIAS;             *A_RES_ALIAS =         \'urn:perun:resource:attribute-def:def:o365ResourceAlias';
+our $A_USER_LOGIN;            *A_USER_LOGIN=             \'urn:perun:user:attribute-def:def:login-namespace:mu';
+
+my @teams = ();
+foreach my $resourceId ($data->getResourceIds()) {
+		my $resAlias = data->getResourceAttributeValue(resource=>$resourceId, attrName=>$A_RES_ALIAS);
```
zlamalp, line 24: `data` -> `$data`

gen/teams_provisioning
```diff
@@ -0,0 +1,43 @@
...
+        print $resAlias . "\n";
+        my $team->{"identity"} = $resAlias;
+        my @members = ();
+        foreach my $memberId ($data->getMemberIdsForResource($resourceId)) {
```
zlamalp, line 28: Method parameters are key-value pairs: `$resourceId` -> `resource => $resourceId`

gen/teams_provisioning
```diff
@@ -0,0 +1,43 @@
...
+			my $userLogin = $data->getMemberAttributeValue(member=> $memberId, attrName=>$A_USER_LOGIN);
```
zlamalp, line 29: This is user attribute, you need to use `getUserAttributeValue()`

gen/teams_provisioning
```diff
@@ -0,0 +1,43 @@
...
+            print $userLogin . "\n";
+            push @members, $userLogin;
+        };
+        $team->{"members"} = \@members;
```
zlamalp, line 33: I suggest to sort members in the list in order to get comparable outputs between script versions/runs. You can add following above this line: `@members = sort(@members);`.

zlamalp (changes requested):

gen/elerning_mu_ff
```diff
@@ -0,0 +1,63 @@
...
+
+my $fileName = "$DIRECTORY/$::SERVICE_NAME";
+
+my $users;
+#####################################
+
+foreach my $resourceId ($data->getResourceIds()) {
+	foreach my $memberId ($data->getMemberIdsForResource(resource => $resourceId)) {
+		my $memberStatus = $data->getMemberAttributeValue(member=> $memberId, attrName=>$A_USER_STATUS);
+		my $memberLogin = $data->getMemberAttributeValue(member=> $memberId, attrName=>$A_USER_FACILITY_LOGIN);
```
zlamalp, line 33: This should be also user attribute, so use `getUserAttributeValue()`. And probably change variable `$A_USER_FACILITY_LOGIN` to `$A_USER_LOGIN`, since its not bound to virtual user-facility attribute, but defined user attribute from specific namespace (including its definition above).

zlamalp (changes requested):

gen/elerning_mu_ff
```diff
@@ -0,0 +1,63 @@
...
+our $SERVICE_NAME = basename($0);
+our $PROTOCOL_VERSION = "3.0.1";
+my $SCRIPT_VERSION = "3.0.2";
+
+perunServicesInit::init;
+my $DIRECTORY = perunServicesInit::getDirectory;
+my $data = perunServicesInit::getHashedHierarchicalData;
+
+#Constants
+our $A_USER_LOGIN;          *A_USER_FACILITY_LOGIN =            \'urn:perun:user:attribute-def:def:login-namespace:mu';
```
zlamalp, line 17: Names must match between `$A_USER_LOGIN` and `*A_USER_FACILITY_LOGIN`. Also please align it in columns with other definitions below.

zlamalp (changes requested):

zlamalp (approved):

stavamichal (approved):

PR #678, opened 2022-04-11
feat(insight_mu): new service for Jira insight

xflord:


gen/insight_mu
```diff
@@ -0,0 +1,141 @@
...
+				EmailAddress        => $uco . "\@muni.cz",
+				Expired 			=> "FALSE",
+				LastImportWriteTime => $writeTime
+			};
+			$memberDataById->{$memberId} = $person;
+		}
+		# decide whether group or workplace depending on defined VAZPR attribute
+		if ( ! defined $groupVAZPR || $groupVAZPR eq '' ) {
+			my $group = {
+				Name                => $groupName,
```
Johaney-s, line 77: I think we want groupId in brackets after the name

gen/insight_mu
```diff
@@ -0,0 +1,141 @@
...
+my $workplaceData = {
+	data   => {Contact => [{Group => [{PerunWorkplace => \@workplaceValues}]}]}
+};
+my $personData = {
+	data   => {Contact => [{Person => \@personValues }]}
+};
+
+# process files
+my $directoryName = "Data";
+mkdir $DIRECTORY . "/$directoryName";
```
Johaney-s, line 123: will this work for existing directories? or is the directory removed before?

send/insight_mu
```diff
@@ -0,0 +1,118 @@
...
+fi
+
+IMPORT_TOKEN="`cat /etc/perun/services/insight_mu/$DESTINATION.t`"
+if [ -z "$IMPORT_TOKEN" ]; then
+	echo "No token in file" >&2
+	exit 246
+fi
+
+#call perl script with mandatory options
+$EXECSCRIPT -t $IMPORT_TOKEN -d $TMP_HOSTNAME_DIR
```
stavamichal, line 110: Send the token as a file path instead of as a value.

send/insight_mu
```diff
@@ -0,0 +1,118 @@
...
+#test if exists perl script to call from this one
+if [ ! -f "$EXECSCRIPT" ]; then
+	echo "Can't locate connector file!" >&2
+	exit 253
+fi
+
+IMPORT_TOKEN="`cat /etc/perun/services/insight_mu/$DESTINATION.t`"
+if [ -z "$IMPORT_TOKEN" ]; then
+	echo "No token in file" >&2
+	exit 246
```
stavamichal, line 106: You already have this exit code above in the lock file. Use a different one and unique.

send/insight-connector.pl
```diff
@@ -0,0 +1,255 @@
...
+$GET_EXECUTION_STATUS = $import_links->{'links'}->{'getExecutionStatus'};
+importFilesFromDirectory();
+
+# Finish import
+callServer($SUBMIT_RESULTS_URL, $TYPE_POST, $finalChunk);
+
+my $jsonResult;
+my $counter = 0;
+while ((!$jsonResult || !$jsonResult->{'status'} || $jsonResult->{'status'} ne "DONE") && $counter < $maxwait) {
+	if ($counter >= $maxwait) { die "Timeout for execution status, did not receive import status!\n" }
```
stavamichal, line 93: This line should be processed after the while cycle.

stavamichal (changes requested):

send/insight_mu
Johaney-s: done

Johaney-s: done

send/insight_mu
Johaney-s: fixed and used values below 255

Johaney-s: fixed and used values below 255

send/insight-connector.pl
Johaney-s: i moved the counter before the timeout check instead

Johaney-s:
Increased timeout limit for positive response from execution status endpoint to ~15 mins

send/insight-connector.pl
stavamichal: A little weird is that if the max wait would be 1, it would not even try at all. I would expect this behavior from the value "0".

gen/insight_mu
```diff
@@ -0,0 +1,141 @@
...
+			if ( defined $o365groupName && ! $o365groupName eq '' ) {
+				$groupName = $o365groupName;
+			}
+			my $group = {
+				Name                => $groupName . " (" . $groupCISPR . "/" . $groupId . ")",
+				Description         => $groupDesc,
+				InetVAZPR           => $groupVAZPR,
+				InetCISPR           => $groupCISPR,
+				ExternalObjectID    => $groupId,
+				EmailAddress        => $o365emails[0][0],
```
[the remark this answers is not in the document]
stavamichal: The reason is simple. You are getting a reference to the array, not the array itself. What you can do is something like this:
`my @array = $getArrayFromPerun() ? @($getArrayFromPerun()) : ();`
It will check if there is something in the reference and if yes, you will retype it to the array. If not, you will return an empty array on your own. Then You can use "$o365emails[0]" properly.

stavamichal (approved):

PR #679, opened 2022-04-12
fix(denbi_portal_compute_center): fixed generating project memberships

balcirakpeter:
Project membership was generated through a virtual attribute which is not correct. Therefore, it was
changed so it is generated through the data accuired by standard way.

BREAKING CHANGE: The prefferedMail attribute needs to be added to the service definition.

### DIFF
--- a/gen/denbi_portal_compute_center
+++ b/gen/denbi_portal_compute_center
@@ -22,12 +22,15 @@ my $data = perunServicesInit::getHashedDataWithGroups;
 sub processUsers;
 sub processGroups;
 sub processMemberships;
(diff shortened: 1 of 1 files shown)


zlamalp (commented):
I think this is mostly ok, you should probably update script version and note also other changes in required attributes in breaking change notice (added `user:def:login-namespace:elixir` / removed `group:virt:denbiProjectMembers`).

I would probably remove constant definition for `group:virt:denbiProjectMembers` as it is used only once for backwards compatibility and its not required attribute anymore.

Commented line with previous way of attribute processing is probably not necessary.

PR #681, opened 2022-04-20
fix(openstack_projects): correctly use EPPN attribute to include all identifiers

xflord:


gen/openstack_projects
```diff
@@ -17,7 +17,7 @@ my $data = perunServicesInit::getHashedHierarchicalData;
 our $A_FACILITY_PROJECT_NAMESPACE; *A_FACILITY_PROJECT_NAMESPACE = \'urn:perun:facility:attribute-def:def:projectNamespace';
 our $A_USER_MAIL;                  *A_USER_MAIL =                  \'urn:perun:user:attribute-def:def:preferredMail';
 our $A_USER_FACILITY_LOGIN;        *A_USER_FACILITY_LOGIN =        \'urn:perun:user_facility:attribute-def:virt:login';
-our $A_USER_OPTIONAL_LOGIN;        *A_USER_OPTIONAL_LOGIN =        \'urn:perun:user:attribute-def:virt:optionalLogin-namespace:mu';
+our $A_USER_OPTIONAL_LOGIN;        *A_USER_OPTIONAL_LOGIN =        \'urn:perun:user:attribute-def:virt:eduPersonPrincipalNames';
```
Johaney-s, line 20: please change the variable name, otherwise looks good

Johaney-s (approved):
we plan to discuss if changes of required attributes should be marked with breaking change, it would be case of this pr also.

cuadradek (approved):

PR #684, opened 2022-04-28
feat(gen): Manually run consent in gen scripts

mattjoke:
* added parameter to getHashedHierarchicalData
  and getHashedDataWithGroups subs

### DIFF
--- a/gen/perunServicesInit.pm
+++ b/gen/perunServicesInit.pm
@@ -40,6 +40,9 @@ END { if($DIE_AT_END) { die "Died because of warning(s) occur during processing.
 
 my ($agent, $service, $facility, $servicesAgent, $directory, $tmp_directory, $tmp_directory_destination, $getData_directory, $local_data);
 
+# Parameter that enforces consent evaluation
+our $CONSENT_EVAL = 0;
+
 # Prepare directory for file which will be generated
 # Create VERSION file in this directory. This file contains protocol version
 #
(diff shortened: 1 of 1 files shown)


gen/perunServicesInit.pm
```diff
@@ -143,7 +143,7 @@ sub getHashedHierarchicalData {
 	if(defined $local_data) { return $local_data; }
   my $filterExpiredMembers = shift;
   unless($filterExpiredMembers) { $filterExpiredMembers = 0; }
-  my $data = $servicesAgent->getHashedHierarchicalData(service => $service->getId, facility => $facility->getId, filterExpiredMembers => $filterExpiredMembers);
+  my $data = $servicesAgent->getHashedHierarchicalData(service => $service->getId, facility => $facility->getId, filterExpiredMembers => $filterExpiredMembers, consentEval => $ENV{"CONSENT_EVAL"});
```
xflord, line 146: Again, couldn't we just create a global variable, include -c in the GetOptions call at the beginning of the init method to set it to true if needed and use it here?

xflord (changes requested):

gen/perunServicesInit.pm
```diff
@@ -50,7 +53,7 @@ my ($agent, $service, $facility, $servicesAgent, $directory, $tmp_directory, $tm
 sub init {
 
 	my ($facilityId, $facilityName, $local_data_file, $serviceName, $getDataType);
-	GetOptions ("facilityId|f=i" => \$facilityId, "facilityName|F=s" => \$facilityName, "data|d=s" => \$local_data_file, "serviceName|s=s" => \$serviceName, "getDataType|t=s" => \$getDataType) or die;
+	GetOptions ("consent-eval|c"=>\$CONSENT_EVAL,"facilityId|f=i" => \$facilityId, "facilityName|F=s" => \$facilityName, "data|d=s" => \$local_data_file, "serviceName|s=s" => \$serviceName, "getDataType|t=s" => \$getDataType) or die;
```
Johaney-s, line 56:
```suggestion
	GetOptions ("consentEval|c=i"=>\$CONSENT_EVAL,"facilityId|f=i" => \$facilityId, "facilityName|F=s" => \$facilityName, "data|d=s" => \$local_data_file, "serviceName|s=s" => \$serviceName, "getDataType|t=s" => \$getDataType) or die;
```
we can specify the type here to integer and follow the camelcase convention

xflord: Do we really want ‘=i’, since we only want consentEval to be either false (0) or true (1)? If I understand correctly, this would require us always to specify with ‘-c 1’ and allow other values than 0 and 1 as well.

Johaney-s: you're right, sorry, my bad. @mattjoke could you please remove it?

mattjoke: On it

xflord (approved):

Johaney-s (approved):

PR #686, opened 2022-05-03
feat(o365_mu_users_export): new scripts for o365 mu

Johaney-s:
* propagates logins of o365 accounts
* added gen and send scripts


Johaney-s (commented):
**approve **

PR #688, opened 2022-05-11
feat(o365_mu): removing and preserving resource mailboxes

stavamichal:
- Because of technical issues with updating resource mailboxes data, we
were unable to finish migration of last version of this service to
production envrionment. The logic about resource mailboxes should be
removed, but there is a plan to further work on it in the future. For
this reason it was preserved and only removed from main code.
- Main parts of the code were commented and they are not processed any
more.
- All methodes and variables were preserved (but we are not using them)

### DIFF
--- a/gen/o365_mu
+++ b/gen/o365_mu
@@ -16,8 +16,8 @@ sub saveUsersToFile;
 sub saveGroupsToFile;
 
 our $SERVICE_NAME     = "o365_mu";
-our $PROTOCOL_VERSION = "3.0.0";
-my $SCRIPT_VERSION = "3.0.1";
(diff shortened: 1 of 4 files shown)


zlamalp (approved):

cuadradek (commented):
We usually use _feat_ as commit type - I am not sure if _feature_ would be parsed correctly by the semantic release library.

stavamichal:
> We usually use _feat_ as commit type - I am not sure if _feature_ would be parsed correctly by the semantic release library.

It was changed to feat.

cuadradek (approved):

PR #697, opened 2022-06-13
fix(crm_ceitec): Removed unnecessary code

mattjoke:
* Removed unnecessary checks, because the Perun won't send
  EXPIRED members

### DIFF
--- a/gen/crm_ceitec
+++ b/gen/crm_ceitec
@@ -8,7 +8,7 @@ use perunServicesUtils;
 
 local $::SERVICE_NAME = "crm_ceitec";
 local $::PROTOCOL_VERSION = "3.1.0";
-my $SCRIPT_VERSION = "3.0.6";
+my $SCRIPT_VERSION = "3.0.7";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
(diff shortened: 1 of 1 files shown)


gen/crm_ceitec
```diff
@@ -38,16 +37,8 @@ my $userGroups; # $userGroups->{$login}->{$groupName} = 1;
 # FOR EACH RESOURCE
 foreach my $resourceId ( $data->getResourceIds() ) {
 	foreach my $groupId ( $data->getGroupIdsForResource( resource => $resourceId ) ) {
-
+		# EXPIRED members are not sent from Perun
```
Johaney-s, line 40: You can update the script version and I think we should omit this comment, because if someone decides to reconfigure the service in gui, it won't be updated in this here.

Johaney-s (commented):
Please squash the commits

Johaney-s (approved):

zlamalp (changes requested):
I think it should be considered as either `fix` or `refactor` and not `feat` as we are not adding any feature - we are using existing feature in the script. Script is tested and deployed on production, so its OK otherwise and can be merged once commit message is fixed.

PR #700, opened 2022-06-27
feat(umbraco_mu): new service for umbraco in MU

stavamichal:
* exporting information about resources and users' logins

### DIFF
--- a/gen/umbraco_mu
+++ b/gen/umbraco_mu
@@ -1,0 +1,54 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use perunServicesInit;
+use perunServicesUtils;
+use File::Basename;
+use JSON::XS;
+
+local $::SERVICE_NAME = basename($0);
+local $::PROTOCOL_VERSION = "3.0.0";
+my $SCRIPT_VERSION = "3.0.0";
+
+perunServicesInit::init;
+my $DIRECTORY = perunServicesInit::getDirectory;
+my $data = perunServicesInit::getHashedHierarchicalData;
+
+our $A_USER_LOGIN;                  *A_USER_LOGIN =                  \'urn:perun:user_facility:attribute-def:virt:login';
+our $A_RESOURCE_UMBRACO_GROUP_NAME; *A_RESOURCE_UMBRACO_GROUP_NAME = \'urn:perun:resource:attribute-def:def:umbracoGroupName';
+
+my $siteStruc = {};
+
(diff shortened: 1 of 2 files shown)


gen/umbraco_mu
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use perunServicesInit;
+use perunServicesUtils;
+use File::Basename;
+use perunServicesInit;
```
zlamalp, line 8: Duplicate import

gen/umbraco_mu
```diff
@@ -0,0 +1,53 @@
...
+
+my $siteStruc = {};
+
+foreach my $resourceId ($data->getResourceIds()) {
+	my $groupName = $data->getResourceAttributeValue(resource => $resourceId, attrName => $A_RESOURCE_UMBRACO_GROUP_NAME );
+	next unless $groupName;
+
+	foreach my $memberId ($data->getMemberIdsForResource( resource => $resourceId )) {
+		my $login = $data->getUserFacilityAttributeValue( member => $memberId, attrName => $A_USER_LOGIN );
+		$siteStruc->{$groupName}->{$login} = $login;
```
zlamalp, line 30: You can probably just store `1` as a value and not login, since you use only hash keys later.

gen/umbraco_mu
```diff
@@ -0,0 +1,53 @@
...
+	my $record = {};
+	$record->{'groupName'} = $groupName;
+	my @members = ();
+	foreach my $login (sort keys %{$siteStruc->{$groupName}}) {
+		push @members, $login;
+	}
+	$record->{'members'} = \@members;
+	push @printingStruc, $record;
+}
+
```
zlamalp, line 46: You should probably sort also groups by their names in order to get comparable output:
```perl
@printingStruc = sort { $a->{'groupName'} cmp $b->{'groupName'} } @printingStruc;
```

zlamalp (changes requested):

gen/umbraco_mu
stavamichal: Removed.

gen/umbraco_mu
stavamichal: Changed.

gen/umbraco_mu
stavamichal: Added.

zlamalp (approved):

PR #701, opened 2022-06-27
feat(users_emails): new scripts for getting all users mails

stavamichal:
* get all emails from users on the facility (no matter the VO)
* added gen, send and slave scripts


slave/process-users-emails/changelog
```diff
@@ -0,0 +1,5 @@
+perun-slave-process-users-emails (3.1.1) stable; urgency=low
```
zlamalp, line 1: I know it technically doesn't matter, but shouldn't we start with something like `3.0.0` or `1.0.0`?

zlamalp (changes requested):

PR #702, opened 2022-07-04
feat(atlassian_mu): connector for atlassian via SCIM

Johaney-s:
* connects to Atlassian API to fetch data about users and groups
* reads Perun users from passed directory and compares changes
* resolves diff between Perun and Atlassian and updates using SCIM protocol

### DIFF
--- a/send/atlassian_mu
+++ b/send/atlassian_mu
@@ -1,0 +1,109 @@
+#!/bin/bash
+
+SERVICE_NAME="atlassian_mu"
+EXECSCRIPT="./atlassian_mu_process.pl"
+
+FACILITY_NAME=$1
+DESTINATION=$2
+DESTINATION_TYPE=$3
+
+	#Test if destination is in not empty and in the correct format
+if [ -z "$DESTINATION" ]; then
+	echo "Missing Destination argument (Name of configuration file to retrieve endpoint URL and key from)" >&2
+	exit 252
+else
+	if echo "$DESTINATION" | grep "[^.A-Za-z_0-9-]"; then
+		echo "Bad format of destination!" >&2
+		exit 251
+	fi
+fi
+
+#Test if name of facility is not empty
+if [ -z "$FACILITY_NAME" ]; then
+	echo "Missing FacilityName argument" >&2
+	exit 245
+fi
+
+#Basic path to find data from gen service
+SERVICE_FILES_BASE_DIR="`pwd`/../gen/spool"
+SERVICE_FILES_DIR="$SERVICE_FILES_BASE_DIR/$FACILITY_NAME/$SERVICE_NAME"
+#Just safety check. This should not happen
+if [ ! -d "$SERVICE_FILES_DIR" ]; then echo '$SERVICE_FILES_DIR: '$SERVICE_FILES_DIR' is not a directory' >&2 ; exit 1; fi
+
+#Create lock to disallow calling more than once at time (method similar to locks in slave scripts)
+LOCK_DIR=${LOCK_DIR:=/var/lock}
+LOCK_FILE="${LOCK_DIR}/perun-${SERVICE_NAME}-$DESTINATION.lock"
+LOCK_PIDFILE="$LOCK_FILE/pid"
+function create_lock {
+	if mkdir "${LOCK_FILE}"; then
+		trap 'rm -r -f "${LOCK_FILE}"' EXIT
+		echo $$ > "$LOCK_PIDFILE";
+		if [ $? -ne 0 ]; then
+			echo "Can't create lock file." >&2
+			exit 250
+		fi
+	else
+		# lock file exists, check for existence of concurrent process
+		if ps ax | grep "$SERVICE_NAME" | sed 's/^\([0-9]\+\).*/\1/' | grep "\(^\| \)`cat $LOCK_PIDFILE`\( \|$\)"; then
+		# concurrent process is running - this skript must terminate
+			echo "Concurrent process $SERVICE_NAME is running" >&2
+			exit 249
+		else
+		# lock is not valid; it should be deleted
+			rm -r "$LOCK_FILE"
+			if [ $? -ne 0 ]; then
+				echo "Can't remove not valid lock file." >&2
+				exit 248
+			fi
+			echo "Invalid lock file found and deleted: $LOCK_FILE" >&2
+			mkdir "${LOCK_FILE}"
+			if [ $? -ne 0 ]; then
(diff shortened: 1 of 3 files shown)


gen/scim_atlassian_mu
```diff
@@ -0,0 +1,134 @@
...
+
+#Constants
+our $A_USER_ID;                   *A_USER_ID =               \'urn:perun:user:attribute-def:core:id';
+our $A_USER_LOGIN;                *A_USER_LOGIN =            \'urn:perun:user_facility:attribute-def:virt:login';
+our $A_USER_GIVEN_NAME;           *A_USER_GIVEN_NAME =       \'urn:perun:user:attribute-def:core:firstName';
+our $A_USER_FAMILY_NAME;          *A_USER_FAMILY_NAME =      \'urn:perun:user:attribute-def:core:lastName';
+our $A_GROUP_NAME;                *A_GROUP_NAME =            \'urn:perun:group:attribute-def:core:name';
+
+#my $domain = "\@muni.cz";
+my $domain = "\@test4mu.servicedesk.muni.cz";
```
stavamichal, line 32: This is only for testing purposes, so before the merge, it has to be removed.

gen/scim_atlassian_mu
```diff
@@ -0,0 +1,134 @@
...
+
+use strict;
+use warnings;
+use perunServicesInit;
+use perunServicesUtils;
+use open qw/:std :utf8/;
+use JSON::XS;
+use utf8;
+
+local $::SERVICE_NAME = "scim_atlassian_mu";
```
stavamichal, line 11: SCIM is the protocol here so I would prefer to not be part of the service's name. So maybe "atlassian_mu" would be better, what do you think?

gen/scim_atlassian_mu
```diff
@@ -0,0 +1,134 @@
...
+our $A_USER_FAMILY_NAME;          *A_USER_FAMILY_NAME =      \'urn:perun:user:attribute-def:core:lastName';
+our $A_GROUP_NAME;                *A_GROUP_NAME =            \'urn:perun:group:attribute-def:core:name';
+
+#my $domain = "\@muni.cz";
+my $domain = "\@test4mu.servicedesk.muni.cz";
+
+my $userStruc = {};
+my $groupStruc = {};
+my $membershipStruc = {};
+
```
stavamichal, line 37: You can create correct headers there like:
my $emailHeader = "email";
And then use them in all parts of the code in the same way instead of these random hashes below.

gen/scim_atlassian_mu
```diff
@@ -0,0 +1,134 @@
...
+		$userStruc->{$uid}->{$givenName} = $userGivenName;
+	}
+
+	processMemberships $gid, $uid;
+}
+
+## creates structure for groups.scim file
+sub processGroups {
+	my $groupId = shift;
+	my $resourceId = shift;
```
stavamichal, line 112: In these three functions, you are using 2 different ways how to get references on variables. I prefer to choose only one of them and be consistent in the script.

gen/scim_atlassian_mu
```diff
@@ -0,0 +1,134 @@
...
+sub processUsers;
+sub processGroups;
+sub processMemberships;
+
+#Constants
+our $A_USER_ID;                   *A_USER_ID =               \'urn:perun:user:attribute-def:core:id';
+our $A_USER_LOGIN;                *A_USER_LOGIN =            \'urn:perun:user_facility:attribute-def:virt:login';
+our $A_USER_GIVEN_NAME;           *A_USER_GIVEN_NAME =       \'urn:perun:user:attribute-def:core:firstName';
+our $A_USER_FAMILY_NAME;          *A_USER_FAMILY_NAME =      \'urn:perun:user:attribute-def:core:lastName';
+our $A_GROUP_NAME;                *A_GROUP_NAME =            \'urn:perun:group:attribute-def:core:name';
```
stavamichal, line 29: We shouldn't use the names of objects in Perun for the identification of objects in the external system. For this purpose, you can use a new attribute, for example 'resource:def:atlassianGroupName' set on the Resource where groups are assigned.

stavamichal (changes requested):

send/scim_atlassian_mu.pm
```diff
@@ -0,0 +1,112 @@
+#!/bin/bash
```
stavamichal, line 1: The name of this script is not proper. It should be "scim_atlassian_mu". The file extension "pm" means Perl module.

send/scim_atlassian_mu.pm
```diff
@@ -0,0 +1,112 @@
...
+
+SERVICE_NAME="scim_atlassian_mu"
+EXECSCRIPT="./scim-atlassian-connector.pl"
```
stavamichal, line 4: We have a standard for the name convention of Perl execution scripts. In this case, it should be "scim_atlassian_mu.pl".

send/scim_atlassian_mu.pm
```diff
@@ -0,0 +1,112 @@
...
+
+FACILITY_NAME=$1
+DESTINATION=$2
+DESTINATION_TYPE=$3
+
+	#Test if destination is in not empty and in the correct format
```
stavamichal, line 10: The indentation in this script is not right.

stavamichal (changes requested):

gen/scim_atlassian_mu
Johaney-s: good point

gen/scim_atlassian_mu
Johaney-s: fixed and simplified the script to skip resolving each group separately and rather process whole resource instead

send/scim_atlassian_mu.pm
Johaney-s: thanks, fixed

send/atlassian_mu
```diff
@@ -0,0 +1,112 @@
...
+fi
+
+#Test if name of facility is not empty
+if [ -z "$FACILITY_NAME" ]; then
+	echo "Missing FacilityName argument" >&2
+	exit 245
+fi
+
+#Destination is name of instance (name of file with other information like authorization etc.)
+INSTANCE_NAME=$DESTINATION
```
stavamichal, line 28: I think we can skip this and use $DESTINATION on the only place where the instance name has been used.

stavamichal (approved):

Johaney-s:
Added printing message with task report (number of updated/created/removed objects).

mattjoke (approved):

PR #710, opened 2022-08-02
feat: use generic_send.py in rt_data_vocesnet and users_emails

zlamalp:
- Use generic_send python script instead of old bash script to improve
  security. Switched services are rt_data_vocesnet and users_emails.


zlamalp:
Tested on production. Merge only after #706.

PR #720, opened 2022-08-15
refactor(o365_mu): Processing bulk of entities

stavamichal:
- send script connector was merged with the process script to the one
working script (there is no need for multi-thread application anymore)
- new o365_mu_process script is now completing all logic of the client
site of the service. It is sending several bulks (every bulk has
a predefined number of entities inside) and processing every bulk on the
server site separately. When bulk is processed on the sever side, the
client will resolve it and save new state of objects to the cache.
The script would end with an error if any object was processed
unsuccessfully.


stavamichal:
Do not merge!

send/o365_mu_process.pl
```diff
@@ -202,352 +152,331 @@
...
+					if($DEBUG>0) { print "+++ USER: $key\n"; }
+				}
+			} else {
+				#if user was not processed at all, it is also an error, do not add to the cache
+				print "WARNING - User with ID $key was timeouted (not processed).\n";
+				$failedUsers++;
+				$returnCode = 1;
+			}
+		}
+		#save only if this is hard run (not the DRY run)
```
Johaney-s, line 270: this will never be a dry run due to the condition `if($countOfUsersToProcess > 0 && $DRY_RUN == 0)`

Johaney-s (approved):

mattjoke (approved):

PR #723, opened 2022-08-18
fix(fs_home): shorten window for race-condition

zlamalp:
- Re-check that home folder doesn't exist before mv command
  to prevent race-condition when service is propagated to
  two facilities with shared /home volume.


zlamalp:
I'm closing this as per statement of our colleague it wouldn't make much difference, especially on the NFS. Facility manager needs to know, if volume is shared from different machine and if it is managed by perun or not.

PR #725, opened 2022-08-25
feat: support IS when pushing data to IFIS at VSUP

zlamalp:
- Users relations (studies) can be also fetched from IS.
- Load IS relations in vsup_ifis instead of KOS.
- Use appendParam() to help remove duplicate code.


zlamalp:
Replaced with #744 , non-conflicting parts were merged.

PR #736, opened 2022-11-10
feat(banned_facility_users): new service for returning facility bans

Johaney-s:
* service returns user bans on facility

### DIFF
--- a/gen/banned_facility_users
+++ b/gen/banned_facility_users
@@ -1,0 +1,45 @@
+#!/usr/bin/perl
+use strict;
+use warnings;
+use perunServicesInit;
+use perunServicesUtils;
+use Perun::Agent;
+use JSON::XS;
+
+our $SERVICE_NAME = "banned_facility_users";
+our $PROTOCOL_VERSION = "3.0.0";
+my $SCRIPT_VERSION = "3.0.0";
+
+perunServicesInit::init;
+my $DIRECTORY = perunServicesInit::getDirectory;
+my $data = perunServicesInit::getHashedHierarchicalData;
+
+my $agent = Perun::Agent->new();
+my $bansAgent = $agent->getBanOnFacilityAgent;
+
+#Constants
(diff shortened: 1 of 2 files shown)


mattjoke (approved):

gen/banned_facility_users
```diff
@@ -0,0 +1,45 @@
...
+my %bansStruc;
+my $facilityId = perunServicesInit::getFacility->getId;
+my @bans = $bansAgent->getBansForFacility(facilityId => $facilityId);
+
+foreach my $memberId ($data->getMemberIdsForFacility) {
+	my $isBanned = $data->getUserFacilityAttributeValue( member => $memberId, attrName => $A_U_F_IS_BANNED );
+	if ($isBanned) {
+		my $uid = $data->getUserIdForMember(member => $memberId);
+		foreach my $ban (@bans) {
+			if ($ban->getUserId() eq $uid) {
```
xflord, line 32: `getBan(PerunSession sess, int userId, int faclityId)` exists in FacilityManager, maybe we could add it to bansAgent and then use it here?

send/banned_facility_users
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+export SERVICE_NAME="banned_facility_users"
+
+python3 generic_send.py "$1" "$2" "$3"
```
xflord, line 4: any particular reason python version of the send script is used instead of perl?

gen/banned_facility_users
Johaney-s: I 'd say iterating all facility bans will be better because we decrease the number of rpc calls.

send/banned_facility_users
Johaney-s: i think there's continuous work in progress by @zlamalp on turning all send scripts using generic perl send to use generic python send. security issues i believe?

gen/banned_facility_users
xflord: I see, you’re right.

xflord (approved):

PR #742, opened 2022-11-25
refactor(vsup_ifis): use appendParam() when constructing complex SQLs

zlamalp:
- Simplified code when constructing complex SQLs by using shared appendParam() function.


zlamalp:
Tested in production, works.

PR #744, opened 2022-11-25
feat(vsup_ifis): switch to IS as a study system

zlamalp:
- Load student relations from new source (IS).
- Renamed variables "kos" to "is" and updated comments.


zlamalp:
Prepared for the migration day.

PR #747, opened 2022-11-28
fix(vsup_web): chomp on data for CAO and EXT relations

zlamalp:
- Chomp on data to remove unnecessary newlines for CAO and EXT relations.
- Added appendParam() method to simplify code when constructing complex SQLs.


zlamalp:
Tested and deployed in production - merging.

PR #754, opened 2022-12-27
refactor(drupal_elixir): use only warning when duplicate users are found

xflord:
* with duplicate users the slave script needs to end with an ok return code in order to prevent rerun of the propagation with the same data

### DIFF
--- a/slave/base/bin/perun
+++ b/slave/base/bin/perun
@@ -131,6 +131,17 @@ function log_msg_without_exit {
 	fi
 }
 
+### Log message to stderr and exit with 0 return code
+function log_warn_to_err_exit {
+	TEXT="$1"
+	TIME=`date "+%H:%M:%S"`
+
+	MSG="Info: ${TEXT}"
+	echo "${TIME} ${MSG}" >&2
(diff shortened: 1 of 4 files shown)


zlamalp (changes requested):
Please update also `changelog` file with a new version for both packages.

slave/base/bin/perun
```diff
@@ -131,6 +131,17 @@ function log_msg_without_exit {
...
 
+### Log message to stderr and exit with 0 return code
+function log_warn_to_err_exit {
+	TEXT="$1"
+	TIME=`date "+%H:%M:%S"`
+
+	MSG="Info: ${TEXT}"
+	echo "${TIME} ${MSG}" >&2
+	logger -t "${NAME}" -p daemon.error "${SERVICE}: ${MSG}" &>/dev/null
+	exit 0;
```
zlamalp, line 142: I think that `;` is not necessary here, as there is only a single command on the line. `;` separates multiple command written on one line.

zlamalp (approved):

PR #759, opened 2023-02-01
feat(atlassian_mu): allow changing domain via facility attribute

Johaney-s:
* domain is used in email addresses which uniquely identifies user in atlassian
* now facilityadmins will be able to change the domain
* result mail consists of user_facility login + '@' + facility domain
* fixed missing slashes '/' in api calls
* added total members and users check due to atlassian limits
* group memberships are sent in chunks due to atlassian limits


Johaney-s:
todo:
- atlassian REAL limits for total users and members
- test once again on idm-test

xkostka2 (approved):

PR #760, opened 2023-02-01
fix: changes to send scripts from vsup

xkostka2:
* send scripts ad_group_vsup, ad_user_vsup, ad_user_vsup_service and vsup_tritius have been updated based on bugfixes made on VSUP instance

### DIFF
--- a/send/VsupIfis.pm
+++ b/send/VsupIfis.pm
@@ -14,7 +14,7 @@ use DBI;
 sub init_config($) {
 
 	my $filename = shift;
-	my $configPath = "/home/perun/perun-sync/$filename";
+	my $configPath = "/etc/perun/ifis/$filename";
 	my $result = {};
 
 	open FILE, $configPath or die "Could not open config file $configPath: $!";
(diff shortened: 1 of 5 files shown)


send/VsupIfis.pm
```diff
@@ -14,7 +14,7 @@ use DBI;
 sub init_config($) {
 
 	my $filename = shift;
-	my $configPath = "/home/perun/perun-sync/$filename";
+	my $configPath = "/etc/perun/ifis/$filename";
```
zlamalp, line 17: Is this config used by more services/scrips? If so, then its ok. Otherwise I think it should be in /etc/perun/services/vsup_ifis/$filename as config related to the service.

send/ad_group_vsup
```diff
@@ -79,7 +79,7 @@ print "Group failed to update (members): " . $counter_group_members_not_updated.
 
 $lock->unlock();
 
-if ($counter_group_members_updated_with_errors > 0) { die "Failed to process: " . $counter_group_members_updated_with_errors . " entries.\nSee log at: ~/send/logs/$service_name.log";}
+if ($counter_group_members_updated_with_errors > 0) { die "Failed to process: " . $counter_group_members_updated_with_errors . " entries.\nSee log at: ~/perun-engine/send/logs/$service_name.log";}
```
zlamalp, line 82: When I think about it, logs on instance managed by ansible will be probably in a different location, so the path should be different. This apply to other scripts too.

zlamalp (changes requested):

send/VsupIfis.pm
xkostka2: this config is also used by cron jobs so I guess it is ok this way

zlamalp (approved):

PR #761, opened 2023-02-01
fix: Fix bbmri_collections transport command

dBucik:
force bbmri_collections to use HTTP1.1.


dBucik:
Using other HTTP versions causes server to respond with failure (probably caused by too large request body).

zlamalp (approved):

PR #765, opened 2023-02-20
feat(gen): redundant code removed

Kuliak:
* getFlatData, getHierarchicalData, getDataWithGroups, getDataWithVos, dataToAttributesHashes subroutines removed
* related comments updated
* default getDataType changed from 'Hierarchical' to 'HashedHierarchical'
* related output (help) text adjusted

breaking change: getFlatData, getHierarchicalData, getDataWithGroups, getDataWithVos, dataToAttributesHashes subroutines cannot be used anymore

### DIFF
--- a/gen/pbsmon_json
+++ b/gen/pbsmon_json
@@ -43,6 +43,21 @@ $struc->{'reserved_machines'} = [];
 $struc->{'frontends'} = [];
 
 
+# Get attribute values from obtained attributes (input).
+#
+# Usage:
+# my $data = $attributesAgent->getRequiredAttributes(service => $service->getId, host => $host->getId);
+# my %facilityAttributes = attributesToHash $data
+# print $facilityAttributes{"urn:perun:facility:attribute-def:core:name"}, "\n";
+#
+sub attributesToHash {
+	my %attributesHash;
+	foreach my $attr (@_) {
+		$attributesHash{$attr->getName}=$attr->getValue;
+	}
+	return %attributesHash;
+}
+
 my $physicalMachinesStruc = {};
 
 my $agent = perunServicesInit->getAgent;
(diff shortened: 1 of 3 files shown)


gen/perunServicesUtils.pm
```diff
@@ -12,10 +12,10 @@ no if $] >= 5.018, warnings => "experimental"; #supress warnings on experimetal
 
 #use Perun::Agent;
 
-# Get attributes from getData function and convert them into hash
+# Get attributes from getHashed function and convert them into hash
 #
 # Usage:
-# my $data = perunServicesInit::getHierarchicalData;
+# my $data = perunServicesInit::getHashedHierarchicalData;
 # my %facilityAttributes = attributesToHash $data->getAttributes
```
zlamalp, line 19: I think that this "usage" in docs is not relevant anymore, since `getHashedHierarchicalData()` returns different structure and there is no `getAttributes()` method for it. We might probably think about removal of `attributesToHash()` and `dataToAttributesHashes()` while the first one seems to be used from a single place within the services.

Kuliak: As for `dataToAttributesHashes`, we might say it is not needed anymore, so I deleted it. 
However, as you mentioned, `attributesToHash` is used in `pbsmon_json service`, so it depends whether given service is used (needed) or can be deleted as a whole.

xflord (approved):

gen/pbsmon_json
```diff
@@ -43,6 +43,21 @@ $struc->{'reserved_machines'} = [];
 $struc->{'frontends'} = [];
 
 
+# Get attribute values from obtained attributes (input).
+#
+# Usage:
+# my $data = $attributesAgent->getRequiredAttributes(service => $service->getId, host => $host->getId);;
```
zlamalp, line 49: There should be just a single semicolon at the end.

zlamalp (approved):
Once merged, don't forget to mention changes in `pbsmon_json` gen script in notes for deployment, since its used on instance with manual deployment.

zlamalp (approved):

PR #766, opened 2023-02-28
feat(zabbix_mu): added new service zabbix_mu

xflord:
* added gen, send, slave scripts for zabbix_mu

### DIFF
--- a/gen/zabbix_mu
+++ b/gen/zabbix_mu
@@ -1,0 +1,81 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use perunServicesInit;
+use perunServicesUtils;
+use JSON::XS;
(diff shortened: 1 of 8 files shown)


Johaney-s (commented):
-  consider adding empty resource with zabbix name to groups structure
-  send and slave scripts are not set as executable, please set them so
-  i can see some example prescripts in conf folder of some slave scripts, not sure it shouldn't be part of this service too?

xflord:
>     * i can see some example prescripts in conf folder of some slave scripts, not sure it shouldn't be part of this service too?

Did not notice, thank you

Johaney-s (approved):

mattjoke (approved):

PR #767, opened 2023-03-09
feat: generic send library

Johaney-s:
* send_lib module provides general methods to be used in send scripts
* integrated send_lib to generic_sender.py, we don't want to affect all services all at once
* rewritten three bash scripts to use generic sender or send library module

### DIFF
--- a/send/generic_sender.py
+++ b/send/generic_sender.py
@@ -1,0 +1,196 @@
+import os
+import sys
+import tempfile
+import send_lib
+import subprocess
+
+
+def process(tar_command, transport_command, destination: str, destination_type: str):
+	"""
+	Process data. Extract files and convert to base64 for host windows destination types.
+	:param tar_command:
+	:param transport_command:
+	:param destination:
+	:param destination_type:
+	:return: (returncode, stdout, stderr)
+	"""
+	process_tar = subprocess.Popen(tar_command, stdout=subprocess.PIPE)
+	if destination_type == send_lib.DESTINATION_TYPE_USER_HOST_WINDOWS_PROXY:
+		# converts stdin to base64 and append single space and "$DESTINATION" at the end of it
+		transformation_process_1 = subprocess.Popen("base64", stdin=process_tar.stdout, stdout=subprocess.PIPE)
+		sed_command = ["sed", "-e", "$s/$/ " + destination + "/g"]
+		transformation_process = subprocess.Popen(sed_command, stdin=transformation_process_1.stdout,
+												  stdout=subprocess.PIPE)
+		transformation_process_1.stdout.close()
+	elif destination_type == send_lib.DESTINATION_TYPE_USER_HOST_WINDOWS:
+		# converts stdin to base64
+		transformation_process = subprocess.Popen("base64", stdin=process_tar.stdout, stdout=subprocess.PIPE)
+	else:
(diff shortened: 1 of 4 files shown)


Johaney-s:
Changes to help us get rid of bash send scripts and rewrite them to Python.

These changes will need to be tested for different use cases. Generic sender is rewritten generic send that uses send_lib methods, to test URL destination type that differs the most I hope to be able to test it on banned facility users service. Tinia is using send script locking, openstack was tested manually on available dev instance. It is not the goal to extract all methods reused in bash send scripts at once, it can be extended in the future.

send/generic_sender.py
```diff
@@ -0,0 +1,200 @@
...
+		# errors will be saved to temp file
+		transport_command = prepare_url_transport_command(temp)
+	elif transport_command is None:
+		transport_command = prepare_ssh_transport_command()
+
+	if destination_type == send_lib.DESTINATION_TYPE_USER_HOST_WINDOWS_PROXY:
+		windows_proxy = send_lib.get_windows_proxy()
+		send_lib.check_windows_proxy_format(windows_proxy)
+		host, hostname, port = send_lib.prepare_destination(destination, destination_type, windows_proxy)
+		transport_command.extend(["-p", port])
```
xflord, line 128: Wouldn't a separate condition that adds the port option if 'port' is defined be better for extensibility (if we added more destination types with a custom port)? Then we could also move the windows proxy logic to send_lib (since all we do here is call functions from send_lib anyway). Also, 'DESTINATION_TYPE_USER_HOST_PORT' uses a custom port and is not accounted for here. Not sure whether this destination type is actually used, though.

send/tinia
```diff
@@ -1,119 +1,35 @@
...
+	send_lib.copy_files_to_directory(service_files_dir, tmp_dir, re.compile(service_name))
 
-#Create lock (same like in slave)
-LOCK_DIR=${LOCK_DIR:=/var/lock}
-LOCK_FILE="${LOCK_DIR}/perunv3-${SERVICE_NAME}-$DBNAME.lock"
-LOCK_PIDFILE="$LOCK_FILE/pid"
+execscript = "./tinia_process.pl"
+send_lib.check_script_file(execscript)
+command = [execscript, "-d", dbname, "-p", tmp_dir, "-s", service_name]
+process = subprocess.Popen(command, stdout=subprocess.PIPE)
```
xflord, line 32: just an idea - couldn't we move all of this logic to send_lib and create a function like 'exec_custom_script(execscript, arguments)'? Skimmed through rest of the send scripts and it seems that this appears in multiple send scripts (insight_mu for example), so I guess we would save a couple lines of code

xflord (commented):
It looks like the refactor should function the same way as before, did not test though

send/generic_sender.py
Johaney-s: good idea! i think i unintentionally changed the usage in the generic_sender and put it to condition for USER_HOST_WINDOWS_PROXY instead of for USER_HOST_PORT

xflord (approved):

Johaney-s:
excluded tinia to separate PR - [https://github.com/CESNET/perun-services/pull/768](url)

mattjoke (approved):
👍

PR #771, opened 2023-03-27
feat: use python send script in bbmri send scripts

xflord:
* rewrote bbmri_networks and bbmri_collections from bash to python using the send_lib library


mattjoke (approved):
From the code, it looks good :+1: Nevertheless, it would also be nice for someone else to test it.

Kuliak (approved):
Looks good.

PR #775, opened 2023-04-05
fix(openstack_projects): change structure of the config file

Johaney-s:
* previously '-u username:password' was loaded from the config file for destination
* this fix allows deciding auth credentials as previously - per destination
* if missing, no auth is sent


xflord (approved):
made the same changes to the bbmri scripts

PR #777, opened 2023-04-13
feat(o365_mu_users_export): include mfaStatus check

xflord:
* the gen script now also retrieves the mfaAttribute and saves its value for each user
* the send processing script now stores the mfa information in a separate table, omitting users without mfa settings. This is skipped if the name of the table is not defined in the configuration file

### DIFF
--- a/gen/o365_mu_users_export
+++ b/gen/o365_mu_users_export
@@ -9,7 +9,7 @@ use utf8;
 
 local $::SERVICE_NAME = "o365_mu_users_export";
 local $::PROTOCOL_VERSION = "3.0.0";
-my $SCRIPT_VERSION = "3.0.0";
+my $SCRIPT_VERSION = "3.0.1";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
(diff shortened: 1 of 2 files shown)


gen/o365_mu_users_export
```diff
@@ -29,29 +30,35 @@ my $validLogins = {};
...
+#4] MFA is enforced (at least partially)
 foreach my $memberId ( $data->getMemberIdsForFacility() ) {
 	my $uco = $data->getUserFacilityAttributeValue( member => $memberId, attrName => $A_UF_LOGIN );
-	#skip all users without UCO
+	# skip all users without UCO
 	next unless $uco;
 	my $o365Status = $data->getUserFacilityAttributeValue( member => $memberId, attrName => $A_UF_O365_STATE );
-	#skip all users with 0 or empty value in status attribute (everything except 0 is OK here)
+	# skip all users with 0 or empty value in status attribute (everything except 0 is OK here)
 	next unless $o365Status;
```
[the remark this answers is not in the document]
xflord: Was not aware of this. Fixed.

gen/o365_mu_users_export
```diff
@@ -29,29 +30,36 @@ my $validLogins = {};
 #3] disableMailForward == true
 #OR
 #3] disableMailForward == false AND mailStoreAndForward == true
+#4] MFA is enforced (at least partially)
```
melanger, line 33: I think now it should be something like
```
2] status of user in o365 is not 0
OR
2] MFA is enforced (at least partially)
3] disableMailForward == true
OR
3] disableMailForward == false AND mailStoreAndForward == true
OR
3] MFA is enforced (at least partially)
```

gen/o365_mu_users_export
```diff
@@ -29,29 +30,36 @@ my $validLogins = {};
...
+	}
+	# skip all users with 0 or empty value in o365 status attribute (everything except 0 is OK here) AND empty mfa status
+	next unless ($o365Status || $mfaStatus ne "none");
 	my $disableMailForward = $data->getUserFacilityAttributeValue( member => $memberId, attrName => $A_UF_DISABLE_O365_MAIL_FORWARD );
 	my $storeAndForward = $data->getUserFacilityAttributeValue( member => $memberId, attrName => $A_UF_O365_STORE_AND_FORWARD );
 	unless( $disableMailForward ) {
-		#skip users with set forward without storing a copy
+		# skip users with set forward without storing a copy
 		next unless $storeAndForward;
 	}
```
melanger, line 50: mfaStatus needs to be computed in, how about:
```perl
unless( $disableMailForward || $mfaStatus ne "none" ) {
    # skip users with set forward without storing a copy
    next unless $storeAndForward;
}
```

send/o365_mu_users_export_process.pl
```diff
@@ -86,7 +103,6 @@
 	if($loginsInTable->{$uco}) {
 		$skipped++;
 	} else {
-		my $insertLogin = $dbh->prepare(qq{INSERT INTO $tableName (uco) VALUES (?)});
 		$insertLogin->execute($uco);
```
melanger, line 106: now this is a problem because `validLogins` contains users who have nonempty mfaStatus but should not be present in the O365 table

send/o365_mu_users_export_process.pl
```diff
@@ -95,20 +111,74 @@
 #remove old logins
 foreach my $uco (sort keys %$loginsInTable) {
 	unless($validLogins->{$uco}) {
```
melanger, line 113: the same problem as with insert

mattjoke (approved):

PR #778, opened 2023-04-19
feat(dhus): dhus service

Johaney-s:
* dhus service sends data to keycloak and database
* slave script calls script which will be set on the destination machine
* all attributes that are set on service are process by gen script


xflord (approved):
looks good

PR #779, opened 2023-04-19
feat: 🎸 Update bbmri_* gen scripts for after BBMRI migration

dBucik:
Change uses the BBMRI User ID extracted from a virtual attribute. The reason for doing so is the change in the IDs of users when they are moved into the LS AAI.

DEPLOY only to LS AAI, when BBMRI is migrated.

### DIFF
--- a/gen/bbmri_collections
+++ b/gen/bbmri_collections
@@ -14,7 +14,7 @@ use utf8;
 
 local $::SERVICE_NAME = "bbmri_collections";
 local $::PROTOCOL_VERSION = "1.0.1";
-my $SCRIPT_VERSION = "1.0.2";
+my $SCRIPT_VERSION = "1.0.3";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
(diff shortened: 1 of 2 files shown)


gen/bbmri_collections
```diff
@@ -71,7 +73,8 @@ foreach my $resourceId ( $data->getResourceIds() ) {
 my @users;
 foreach my $uid (sort keys %$userStruc) {
 	my $user = {};
-	$user->{"id"} = $uid;
+	$user->{"id"} = "$userStruc->{$uid}->{$u_bbmri_user_id}";
+	$user->{"perun_user_id"} = $uid;
```
balcirakpeter, line 77: Why do you need Perun user ID?

gen/bbmri_collections
```diff
@@ -27,6 +27,7 @@ sub processMemberships;
 
 #Constants
 our $A_USER_ID;                   *A_USER_ID =               \'urn:perun:user:attribute-def:core:id';
+our $A_BBMRI_USER_ID;             *A_BBMRI_USER_ID =         \'urn:perun:user:attribute-def:virt:bbmriUserId';
```
Johaney-s, line 30:
```suggestion
our $A_BBMRI_USER_ID;             *A_BBMRI_USER_ID =         \'urn:perun:user:attribute-def:virt:login-namespace:bbmriid-persistent';
```

gen/bbmri_networks
```diff
@@ -27,6 +27,7 @@ sub processMemberships;
 
 #Constants
 our $A_USER_ID;                   *A_USER_ID =               \'urn:perun:user:attribute-def:core:id';
+our $A_BBMRI_USER_ID;             *A_BBMRI_USER_ID =         \'urn:perun:user:attribute-def:virt:bbmriUserId';
```
Johaney-s, line 30:
```suggestion
our $A_BBMRI_USER_ID;             *A_BBMRI_USER_ID =         \'urn:perun:user:attribute-def:virt:login-namespace:bbmriid-persistent';
```

gen/bbmri_collections
Johaney-s: this will be the new attribute name

gen/bbmri_collections
dBucik: Just wanted to let it there for debugging purposes, could be useful for further changes after services are re-connected to LS AAI proper proxy.

Johaney-s (approved):

balcirakpeter (approved):

PR #774, opened 2023-04-04
feat(send): atlassian_mu services in Python

Kuliak:
* atlassian_mu script rewritten into Python using send_lib library
* atlassian_mu_process.pl file is not part of this change

PR #780, opened 2023-04-20
feat(send): atlassian_mu services in Python

xflord:
* atlassian_mu script rewritten into Python using send_lib library
* atlassian_mu_process.pl file is not part of this change

### DIFF
--- a/send/atlassian_mu
+++ b/send/atlassian_mu
@@ -1,109 +1,36 @@
-#!/bin/bash
+#!/usr/bin/env python3
 
-SERVICE_NAME="atlassian_mu"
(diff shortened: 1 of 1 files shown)


xflord:
Differences from https://github.com/CESNET/perun-services/pull/774:

- fixed regular expressions (files didn't copy and script would always fail on argument check)
- properly get tmpDir path
- wait for the process script to finish (otherwise master slave script would finish beforehand and remove the temporary directory)
- also removed the destination_type argument, since it was unused

mattjoke (approved):
Looks good

bodnara (approved):

PR #783, opened 2023-05-02
feat(feudal): rewrite feudal send script to python

xflord:


mattjoke (approved):
Looks good, if tested👍

send/feudal.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python3.7
```
balcirakpeter, line 1:
```suggestion
#!/usr/bin/env python3
```

send/feudal.py
```diff
@@ -0,0 +1,62 @@
...
+if re.fullmatch(re.compile(r'.*\.\..*'), credentials_filename):
+	send_lib.die_with_error("Unsupported destination (\"..\" cannot be part of destination name)")
+
+with open(f'/etc/perun/services/feudal/{credentials_filename}', 'r') as f:
+	credentials = f.read().strip()
+
+if not credentials:
+	send_lib.die_with_error('No credentials available in configuration')
+# TODO verify credentials file format
+auth = (credentials.split(' ')[0], credentials.split(' ')[1])
```
balcirakpeter, line 34:
```suggestion
auth = (credentials.split(':')[0], credentials.split(':')[1])
```

send/feudal.py
```diff
@@ -0,0 +1,62 @@
...
+
+service_files_dir = send_lib.get_gen_folder(facility, SERVICE_NAME)
+
+send_lib.create_lock(SERVICE_NAME, destination)
```
balcirakpeter, line 38: This does not work as the destination contains `//`. Overall we need to escape such characters in the `send_lib.py` so we do not encounter the same problem next time. Can you please add the escaping there?

balcirakpeter (changes requested):

send/feudal.py
balcirakpeter, line 38:
```suggestion
send_lib.create_lock(SERVICE_NAME, re.sub(r'^https?://', '', destination))
```

PR #788, opened 2023-05-28
feat(crocs_user_export): rewrite send script to python

sarkapalkovicova:
Send script for crocs_user_export was rewrote to python.

### DIFF
--- a/send/crocs_user_export
+++ b/send/crocs_user_export
@@ -1,41 +1,22 @@
-#!/bin/bash
+#!/usr/bin/env python3
+import sys
+import send_lib
+from datetime import datetime
+import re
 
-SERVICE_NAME="crocs_user_export"
+SERVICE_NAME = "crocs_user_export"
 
-TIMEOUT="5400" #90s * 60 sec = 1.5h
-TIMEOUT_KILL="60" # 60 sec to kill after timeout
+send_lib.check_input_fields(sys.argv)
 
-FACILITY_NAME=$1
-DESTINATION=$2
-DESTINATION_TYPE=$3
+facility = sys.argv[1]
+destination = sys.argv[2]
+destination_type = send_lib.DESTINATION_TYPE_SERVICE_SPECIFIC if len(sys.argv) < 4 else sys.argv[3]
 
-SERVICE_FILES_BASE_DIR="`pwd`/../gen/spool"
-SERVICE_FILES_DIR="$SERVICE_FILES_BASE_DIR/$FACILITY_NAME/$SERVICE_NAME"
+send_lib.check_destination_type_allowed(destination_type, send_lib.DESTINATION_TYPE_SERVICE_SPECIFIC)
+send_lib.check_destination_format(destination, destination_type)
 
+service_files_dir = send_lib.get_gen_folder(facility, SERVICE_NAME)
 
-#Just safety check. This should not happen.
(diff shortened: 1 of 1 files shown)


send/crocs_user_export.py
```diff
@@ -0,0 +1,32 @@
...
+service_files_dir = send_lib.get_gen_folder(facility, SERVICE_NAME)
+
+#unless specific configuration for destination exists use common configuration for all destination
+service_files_for_destination = service_files_dir + "/_destination/" + destination
+
+if not os.path.isdir(service_files_for_destination):
+	service_files_for_destination = service_files_dir + "/_destination/all"
+
+send_lib.check_destination_type_allowed(destination_type, send_lib.DESTINATION_TYPE_SERVICE_SPECIFIC)
+send_lib.check_destination_format(destination, destination_type)
```
xflord, line 28: I would probably move these checks to right after line 17 (to keep the argument retrieving/validating logic coupled together)

send/crocs_user_export.py
```diff
@@ -0,0 +1,32 @@
...
+
+timestamp = datetime.now().strftime("%Y-%m-%d_%H.%M.%S")
+
+send_lib.copy_files_to_directory(service_files_dir, destination + "/crocs-user-export-" + timestamp, re.compile('^({})$'.format(SERVICE_NAME)))
```
xflord, line 32:
```suggestion
send_lib.copy_files_to_directory(service_files_dir, destination + "/crocs-user-export-" + timestamp, re.compile(SERVICE_NAME))
```

send/crocs_user_export.py
```diff
@@ -0,0 +1,32 @@
...
+send_lib.check_input_fields(sys.argv)
+
+facility = sys.argv[1]
+destination = sys.argv[2]
+destination_type = send_lib.DESTINATION_TYPE_SERVICE_SPECIFIC if len(sys.argv) < 4 else sys.argv[3]
+
+service_files_dir = send_lib.get_gen_folder(facility, SERVICE_NAME)
+
+#unless specific configuration for destination exists use common configuration for all destination
+service_files_for_destination = service_files_dir + "/_destination/" + destination
```
xflord, line 22: what is this variabe for?

send/crocs_user_export.py
```diff
@@ -0,0 +1,32 @@
...
+import sys
+import os.path
+import send_lib
+from datetime import datetime
+import re
+
+SERVICE_NAME = "crocs_user_export"
+
+TIMEOUT = "5400" #90s * 60 sec = 1.5h
+TIMEOUT_KILL = "60" # 60 sec to kill after timeout
```
xflord, line 11: What are these for?

send/crocs_user_export.py
bodnara: I have checked with @zlamalp and it is most likely some artifact of the past and can be removed. However we should check if the gen script does not try to generate files in that location.

bodnara (approved):

sarkapalkovicova:
> Looks good. Also please check with someone whether we want to keep the old send scripts or remove them and replace them with the new python ones. (the .py suffix might cause problems?)

After discussion with @zlamalp, old script was replaced with new Python one.

zlamalp (approved):
Tested in production, works.

PR #789, opened 2023-05-29
fix(send): escape forbidden characters in destination

Kuliak:
* new function added into send_lib library for filename escaping

### DIFF
--- a/send/send_lib.py
+++ b/send/send_lib.py
@@ -265,7 +265,7 @@ def create_lock(service_name: str, destination: str, custom_lock_dir: str = None
 			lock_dir = TEMPORARY_DIR
 
 	if destination is not None:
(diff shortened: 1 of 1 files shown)


send/send_lib.py
```diff
@@ -329,4 +329,22 @@ def exec_script(script_path: str, arguments: list[str]):
...
+
+def escape_filename(filename: str) -> str:
+	"""
+	Checks if provided filename contains any forbidden characters and escapes them.
+
+	:param filename: filename to escape
+	:return: escaped filename
+	"""
+	escaped_filename: str = ""
+	for char in filename:
```
xflord, line 343: Wouldn't the 'replace' method be much simpler than a loop since we're escaping just one character?

bodnara (approved):

xflord (approved):

PR #792, opened 2023-06-12
feat(ftps_generic): new service for explicit ftps file transport

HejdaJakub:
* This service uses generic gen script and then sends generated data (file) via explicit ftps.

### DIFF
--- a/send/ftps_generic
+++ b/send/ftps_generic
@@ -1,0 +1,50 @@
+#!/usr/bin/env python3
+import os
+import sys
+import re
+from ftplib import FTP_TLS
+from urllib.parse import urlparse
+
+import send_lib
+import tarfile
+
+service_name = "ftps_generic"
+port = 21
+
+file_path_pattern = re.compile("[a-zA-Z0-9\+\-~_\./]*")
+
+if __name__ == "__main__":
+	# Parse command line arguments
+	send_lib.check_input_fields(sys.argv)
+	facility_name = sys.argv[1]
+	destination = sys.argv[2]
+
+	# Parse destination
+	# expected destination type url, example: 'ftp://example.org:2121/path-to/folder/'
+	parsed_destination = urlparse(destination, scheme='ftp')
+	if parsed_destination.scheme not in ['ftp', 'ftps', 'ftpes']:
+		print("only FTPS is supported by this service")
+		exit(1)
+	hostname = parsed_destination.hostname
+	port = parsed_destination.port or port
+	remote_folder_path = parsed_destination.path.rstrip('/')
+
+	local_file_path = os.getcwd() + "/../gen/spool/" + facility_name + "/" + service_name + "/" + service_name
+	remote_file_path = f'{remote_folder_path}/{service_name}'
+	if not (re.fullmatch(file_path_pattern, remote_file_path)):
+		send_lib.die_with_error("Path to the remote file " + remote_file_path + "  is not in a valid format")
+
+	# Connect via explicit FTPS (FTPES)
+	auth = send_lib.get_auth_credentials(service_name, destination)
+	ftps = FTP_TLS()
+	ftps.connect(hostname, port)
+	ftps.login(auth[0], auth[1])
+	ftps.prot_p()
+
+	# Send file to FTP server
+	with open(local_file_path, 'rb') as file:
+		ftps.storbinary(f'STOR {remote_file_path}', file)
+
+	# End connection
+	print("File has been successfully sent")
+	ftps.quit()
--- a/send/send_lib.py
+++ b/send/send_lib.py
@@ -340,3 +340,33 @@ def escape_filename(filename: str) -> str:
 	:return: escaped filename
 	"""
 	return filename.replace('/', '⧸')  # replace classic slash with U+29F8
+
+
+def get_auth_credentials(service_name: str, destination: str):
+	"""
+	Retrieves credentials from a file located in the /etc/perun/services/service_name/service_name.py file.
+	The file is a python file containing map of {destination: {'username': username, 'password': password}} mapping.
+
+	Example of such file - supplement $ variables with destination url, username and password
+	credentials = {
+		"$url1": { 'username': "$user1", 'password': "$pwd1" },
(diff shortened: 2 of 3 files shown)


HejdaJakub:
Credentials are available on devel machine `/etc/perun/services/ftps_generic/auth`.

Johaney-s (changes requested):

send/ftps_generic
```diff
@@ -0,0 +1,53 @@
...
+	if len(sys.argv) == 3:
+		# if there is no remote location, use default "/" location
+		remote_folder_path = '/'
+	else:
+		remote_folder_path = sys.argv[3]
+
+	local_file_path = os.getcwd() + "/../gen/spool/" + facility_name + "/" + service_name + "/" + service_name
+	remote_file_path = f'{remote_folder_path}/{service_name}'
+
+	# Store username and password
```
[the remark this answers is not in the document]
Johaney-s: also, this approach won't handle different auth credentials for different destinations - you can do it same as in bbmri_collections - use destination as a key in a python file and import it to the send script

send/ftps_generic
```diff
@@ -0,0 +1,53 @@
...
+		print("Error: Expected number of arguments is 2 or 3 (FACILITY_NAME, DESTINATION and optional REMOTE_LOCATION)", file=sys.stderr)
+		exit(1)
+	
+	facility_name = sys.argv[1]
+	destination = sys.argv[2]
+	if len(sys.argv) == 3:
+		# if there is no remote location, use default "/" location
+		remote_folder_path = '/'
+	else:
+		remote_folder_path = sys.argv[3]
```
melanger, line 21:
```suggestion
	destination = sys.argv[2]
	# expected destination type url, example: 'ftp://example.org:2121/path-to/folder/'
	parsed_destination = urllib.parse.urlparse(destination, scheme='ftp')
	if parsed_destination['scheme'] not in ['ftp', 'ftps', 'ftpes']:
	    # TODO: print("only FTPS is supported by this service"); exit();
	destination = parsed_destination['hostname']
	port = parsed_destination['port'] or port
	remote_folder_path = parsed_destination['path'].rstrip('/') + '/'
```

melanger: it should throw a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) when an invalid URL is specified, which you can catch or leave, depending on the convention used by other services

send/send_lib.py
```diff
@@ -340,3 +340,34 @@ def escape_filename(filename: str) -> str:
...
+
+	try:
+		sys.path.insert(1, f'{SERVICES_DIR}/{service_name}/')
+		credentials = __import__(service_name).credentials
+		if destination in credentials.keys():
+			username = credentials.get(destination).get('username')
+			password = credentials.get(destination).get('password')
+			auth = (username, password)
+	except Exception:
+		print("Error")
```
Johaney-s, line 370:
```suggestion
```

Johaney-s (approved):

send/ftps_generic
```diff
@@ -0,0 +1,45 @@
...
+	# Connect via explicit FTPS (FTPES)
+	auth = send_lib.get_auth_credentials(service_name, destination)
+	ftps = FTP_TLS()
+	ftps.connect(hostname, port)
+	ftps.login(auth[0], auth[1])
+	ftps.prot_p()
+
+	# Send file to FTP server
+	with open(local_file_path, 'rb') as file:
+		ftps.storbinary(f'STOR {remote_file_path}', file)
```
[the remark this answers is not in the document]
melanger:
```suggestion
        remote_file_path = remote_file_path.replace('\\', '\\\\').replace('"', '\\"')
		ftps.storbinary(f'STOR "{remote_file_path}"', file)
```

if it can be assumed that no control characters (like newline) will be present in the destination string

melanger (changes requested):

send/ftps_generic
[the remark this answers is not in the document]
HejdaJakub: I added the replacement, but I removed the quotas around `"{remote_file_path}"`, without them it works correctly I think.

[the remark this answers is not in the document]
melanger: The quotes are needed for it to be safe. If it does not work with the server, the validation needs to be changed to whitelisting a regex, something like `\w`

send/ftps_generic
```diff
@@ -0,0 +1,50 @@
...
+from ftplib import FTP_TLS
+from urllib.parse import urlparse
+
+import send_lib
+import tarfile
+
+service_name = "ftps_generic"
+port = 21
+
+file_path_pattern = re.compile("[a-zA-Z0-9+\-\/&@#%?=~_|!:,\.;()*$']*")
```
melanger, line 14: I was thinking and I would be more strict, because nothing else is needed
```suggestion
file_path_pattern = re.compile("[a-zA-Z0-9\+\-~_\.]*")
```

send/ftps_generic
```diff
@@ -0,0 +1,50 @@
...
+from ftplib import FTP_TLS
+from urllib.parse import urlparse
+
+import send_lib
+import tarfile
+
+service_name = "ftps_generic"
+port = 21
+
+file_path_pattern = re.compile("[a-zA-Z0-9\+\-~_\.]*")
```
melanger, line 14: Sorry, read the code wrong, it applies to path, not only file name.
```suggestion
file_path_pattern = re.compile("[a-zA-Z0-9\+\-~_\./]*")
```

send/ftps_generic
```diff
@@ -0,0 +1,50 @@
...
+
+	# Connect via explicit FTPS (FTPES)
+	auth = send_lib.get_auth_credentials(service_name, destination)
+	ftps = FTP_TLS()
+	ftps.connect(hostname, port)
+	ftps.login(auth[0], auth[1])
+	ftps.prot_p()
+
+	# Send file to FTP server
+	if not (re.fullmatch(file_path_pattern, remote_file_path)):
```
melanger, line 43: I would put this before auth

HejdaJakub: I've moved it before auth.

melanger (approved):

PR #795, opened 2023-06-16
feat(umbraco_mu): change send script to use BA credentials

Johaney-s:


HejdaJakub (approved):
The code looks good :+1:

PR #802, opened 2023-07-03
feat(tinia): rename externists and add gotex chips

martin-kuba:
Update from version set by Ansible in idm.ics.muni.cz/gen-local - adds marking of externists and
adding chip numbers from attribute gotexChipNumbers.

DEPLOYMENT NOTE: remove gen-local/tinia, it is the same now

### DIFF
--- a/gen/tinia
+++ b/gen/tinia
@@ -9,7 +9,7 @@ use utf8;
 
 local $::SERVICE_NAME = "tinia";
 local $::PROTOCOL_VERSION = "3.0.0";
-my $SCRIPT_VERSION = "3.0.2";
+my $SCRIPT_VERSION = "3.0.3";
 
 perunServicesInit::init;
 my $DIRECTORY = perunServicesInit::getDirectory;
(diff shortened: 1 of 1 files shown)


gen/tinia
```diff
@@ -47,21 +48,28 @@ foreach my $resourceId ( $data->getResourceIds() ) {
 	my $resourceArming = $data->getResourceAttributeValue( resource => $resourceId, attrName => $A_RESOURCE_ARMING );
 	my $resourceAccGroup = $data->getResourceAttributeValue( resource => $resourceId, attrName => $A_RESOURCE_ACC_GROUP );
 
+	#skip resource with COVID-APPROVAL (for now)
+	if( $resourceAccGroup eq 'COVID-APPROVAL' ) { next; }
```
Johaney-s, line 52: is there a reason to have these conditions here? (instead of unassigning the resources from the service)?

gen/tinia
```diff
@@ -47,21 +48,28 @@ foreach my $resourceId ( $data->getResourceIds() ) {
...
-		my $type = "1";
-
-		foreach my $chipNumber (@{$chipNumbers}) {
+		my $type = $external ? "2" : "1";
+		my @allChipNumbers = (@{$chipNumbers // []}, @{$gotexChipNumbers // []});
+		
+		foreach my $chipNumber (@allChipNumbers) {
 			if(!defined($dataByChip->{$chipNumber})) {
-				$dataByChip->{$chipNumber}->{$FIRSTNAME} = $firstName;
+				$dataByChip->{$chipNumber}->{$FIRSTNAME} = $external ? 'EXTERNISTA' : $firstName;
```
Johaney-s, line 72: from code point of view it's ok, cannot speak for how it affects the service

Johaney-s:
please change the commit header to something like `feat(tinia): rename externists and update gotex chips`

PR #814, opened 2023-08-28
feat(webcentrum_eosc): new configurable service

Johaney-s:
* service is configurable by setting required attributes, generates general data format

### DIFF
--- a/gen/webcentrum_eosc
+++ b/gen/webcentrum_eosc
@@ -1,0 +1,12 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use File::Basename;
+use perunDataGenerator;
+
+local $::SERVICE_NAME = basename($0);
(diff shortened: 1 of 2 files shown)


send/umbraco_mu
```diff
@@ -18,6 +18,7 @@ send_lib.check_destination_format(destination, destination_type)
 
 auth = send_lib.get_auth_credentials(SERVICE_NAME, destination)
 
+send_lib.create_lock(SERVICE_NAME, destination)
```
xflord, line 21: maybe this should be a separate commit (or at least a mention in the message)?

xflord (approved):
looks good

send/umbraco_mu
Johaney-s: added to commie message

PR #816, opened 2023-08-30
feat(pithia_portal): added authorization token support

xflord:
* rewrote pithia_portal send script into python and added token support

### DIFF
--- a/send/pithia_portal
+++ b/send/pithia_portal
@@ -1,4 +1,38 @@
-#!/bin/bash
-export SERVICE_NAME="pithia_portal"
+#!/usr/bin/env python3
+import sys
+import send_lib
+import requests
+import generic_sender
 
-python3 generic_send.py "$1" "$2" "$3"
+SERVICE_NAME = "pithia_portal"
+
+send_lib.check_input_fields(sys.argv, True)
+
+facility = sys.argv[1]
+destination = sys.argv[2]
+destination_type = sys.argv[3]
+
+send_lib.check_destination_type_allowed(destination_type, send_lib.DESTINATION_TYPE_URL)
+
+auth_token = None
+
+auth = send_lib.get_auth_credentials(SERVICE_NAME, destination)
+tokenUrl = send_lib.get_custom_config_properties(SERVICE_NAME, destination, ["tokenUrl"])[0]
+data = {
+	"username": auth[0],
+	"password": auth[1]
+}
+response = requests.post(tokenUrl, data)
+resp_cookies = response.cookies.get_dict()
+if response.status_code == 200:
+	response = response.json()
+	if "token" not in response:
+		send_lib.die_with_error("Token missing in response from token server")
+	auth_token = response["token"]
+else:
+	send_lib.die_with_error("Response from token server ended with error code: " + str(response.status_code)
+							+ ", content: " + response.text)
+
+opts = ['-H', 'Authorization: Bearer ' + auth_token, '--cookie', 'sessionid=' + resp_cookies['sessionid']]
+
+generic_sender.send(SERVICE_NAME, facility, destination, destination_type, opts)


send/pithia_portal.py
```diff
@@ -0,0 +1,39 @@
...
+data = {
+	"username": auth[0],
+	"password": auth[1]
+}
+response = requests.post(url, data)
+
+if response.status_code == 200:
+	response = response.json()
+	if "token" not in response:
+		send_lib.die_with_error("Token missing in response")
```
zlamalp, line 27: `Token missing in response from token server` - so it can't be confused with an error from the actual service endpoint.

send/generic_sender.py
```diff
@@ -83,6 +83,8 @@ def prepare_url_transport_command(temp, content_type: str = "application/x-tar",
 	# add standard CURL params
 	transport_command.extend(["-i", "-H", "Content-Type:" + content_type, "-w", "%{http_code}", "--show-error",
 							  "--silent", "-o", temp.name, "-X", method, "--data-binary", "@-"])
+	if auth_token:
+		transport_command.extend(["-H", "Authorization: Bearer " + auth_token])
```
zlamalp, line 87: Based on the original docs, they are expecting Authorization header just with the token value (no `Bearer`). At least it worked that way when I originally tested the API. I'll check with the service provider.

Note: Just because they are using "tokens" it doesn't have to mean they are implementing OAuth 2.0, which bearer authorization scheme is a part of. But I would prefer to use the standard, so I'll talk to them.

zlamalp (changes requested):

send/generic_sender.py
```diff
@@ -107,6 +109,7 @@ def prepare_ssh_transport_command():
 		if destination_type == send_lib.DESTINATION_TYPE_EMAIL or destination_type == send_lib.DESTINATION_TYPE_SERVICE_SPECIFIC:
 			print("Destination type " + destination_type + " is not supported yet.", file=sys.stderr)
 			exit(1)
+	auth_token = sys.argv[4]
```
zlamalp, line 112: This doesn't work since we check number of input params in `send_lib.check_input_fields(sys.argv)` a few lines above and its implementation needs to be updated for this new optional param.

send/generic_sender.py
zlamalp: They updated the enpoint to expect Bearer tokens, just the verification fails for now.

zlamalp (changes requested):

send/pithia_portal.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
+import os
+import sys
+import subprocess
+import send_lib
+import requests
```
zlamalp, line 6: You will need this for other changes:
```python
import generic_sender
```

send/pithia_portal.py
```diff
@@ -0,0 +1,41 @@
...
+auth_token = None
+
+auth = send_lib.get_auth_credentials(SERVICE_NAME, destination)
+tokenUrl = send_lib.get_custom_config_properties(SERVICE_NAME, destination, ["tokenUrl"])[0]
+data = {
+	"username": auth[0],
+	"password": auth[1]
+}
+response = requests.post(tokenUrl, data)
+
```
zlamalp, line 23: For some reason they need session cookie retrieved from the token endpoint so you can use this:
```python
resp_cookies = response.cookies.get_dict()
```

send/pithia_portal.py
```diff
@@ -0,0 +1,41 @@
...
+if response.status_code == 200:
+	response = response.json()
+	if "token" not in response:
+		send_lib.die_with_error("Token missing in response from token server")
+	auth_token = response["token"]
+else:
+	send_lib.die_with_error("Response from token server ended with error code: " + str(response.status_code)
+							+ ", content: " + response.text)
+
+# TODO update with function `main` function call when merged
```
zlamalp, line 33: As mentioned in TODO, since #822 is merged all code below can be replaced with:
```python
opts = ['-H', 'Authorization: Bearer '+auth_token, '--cookie', 'sessionid='+resp_cookies['sessionid'] ]

generic_sender.send(service_name, facility, destination, destination_type, opts)
```

zlamalp (changes requested):
We shouldn't need any changes outside the service send script now when #822 is merged. Also please remove `.py` from the file name, since its called by engine without it.

send/pithia_portal
```diff
@@ -1,4 +1,42 @@
-#!/bin/bash
-export SERVICE_NAME="pithia_portal"
+#!/usr/bin/env python3
+import os
+import sys
+import subprocess
```
zlamalp, line 4: This import seems to be unused.

send/pithia_portal
```diff
@@ -1,4 +1,42 @@
...
+os.environ["SERVICE_NAME"] = SERVICE_NAME
+
+send_lib.check_input_fields(sys.argv, True)
+
+facility = sys.argv[1]
+destination = sys.argv[2]
+destination_type = sys.argv[3]
+
+send_lib.check_destination_type_allowed(destination_type, send_lib.DESTINATION_TYPE_URL)
+send_lib.check_destination_format(destination, destination_type)
```
zlamalp, line 19: Check on the destination format is part of the `generic_sender.send()` implementation. Check on allowed destination type is a good catch, since it's specific to this service.

send/pithia_portal
```diff
@@ -1,4 +1,42 @@
...
+import os
+import sys
+import subprocess
+import send_lib
+import requests
+import generic_sender
 
-python3 generic_send.py "$1" "$2" "$3"
+SERVICE_NAME = "pithia_portal"
+os.environ["SERVICE_NAME"] = SERVICE_NAME
```
zlamalp, line 10: I don't think we need to export service name as environmental variable anymore as it is passed directly to methods it is used in, and its not used in methods missing it in their API.

zlamalp (changes requested):

zlamalp (approved):
I tested the latest version on production and it works as expected. Thanks for good work.

HejdaJakub (approved):
I read the comments about all the progress and the script looks really clear. Good job.

PR #817, opened 2023-08-30
feat(atlassian_mu): add 'del_' prefix to inactive users

xflord:
* when processing to atlassian, inactive users are now marked with the 'del_' prefix
* an extra function to add the prefix to all currently inactive users was also added, its use is commented (can be removed upon initialization)

### DIFF
--- a/send/atlassian_mu_process.pl
+++ b/send/atlassian_mu_process.pl
@@ -19,6 +19,8 @@ use Array::Utils qw(:all);
 # (0) no debug, (1) only messages of performed operations, (2) communication with server.
 my $DEBUG=0;
 
+my $DEACTIVATED_PREFIX_INIT=0;
+
 # Required setting for connection
 my $RESULTS_COUNT = 200; # This many objects will be fetched with one GET request on server. Limited by Atlassian ANYWAY!
 my $directoryUrl;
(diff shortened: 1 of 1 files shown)


mattjoke (approved):

send/atlassian_mu_process.pl
```diff
@@ -437,7 +461,15 @@ sub updateUser {
 	my $userInfo = shift;
 	my $userId = shift;
 	my $statusActive = shift;
-	if ($DEBUG) {print $statusActive ? "Updating user $userName.\n" : "Deactivating user $userName.\n" ;}
+	my $prefixUpdate = shift;
```
Johaney-s, line 464: not sure if it's part of the temporary solution, but i think this flag is not needed, as it can be resolved by removing the prefix only if it exists, but it's just a cosmetic detail. once the temporary code is removed, it's value will always be false

xflord: I did it mainly for debugging purposes, and since it felt wrong that the update counter increased for users updated with the prefix, but will remove

Johaney-s (approved):
i would keep the temporary changes out of the repo and run them only locally, code looks OK

send/atlassian_mu_process.pl
```diff
@@ -100,9 +100,10 @@ sub help {
 my $ourGroups = fetchOurGroups();
 # Data from Atlassian into hash
 my $theirUsers = fetchTheirUsers();
-my $theirGroups = fetchTheirGroups();
 
+addPrefixToDeactivatedUsers(); # REMOVE THIS LINE AFTER PREFIX INITIALIZATION
```
melanger, line 104:
```suggestion
my $DEACTIVATED_PREFIX_INIT=0;

if ($DEACTIVATED_PREFIX_INIT) {
    addPrefixToDeactivatedUsers();
}
```

+ update DEPLOYMENT NOTE accordingly (`scripts needs to be run once with DEACTIVATED_PREFIX_INIT=1`)

melanger (changes requested):
squash commits

melanger (approved):

PR #822, opened 2023-09-05
feat: extract main generic_sender logic to send() python method

zlamalp:
- Previously all generic processing was part of the `__main__`, expected to be started from the cmd line only. Now it is extracted to the defined `send()` method and can be called directly from python scripts too.
- With this solution, we can still use the script from the command line, when `__main__` simply checks input and calls `send()` instead.
- New `opts` param was added to the `send()`, so the custom send scripts can extend default transport commands for SSH/cURL.

### DIFF
--- a/send/generic_sender.py
+++ b/send/generic_sender.py
@@ -95,21 +95,25 @@ def prepare_ssh_transport_command():
 			"GSSAPIAuthentication=no", "-o", "GSSAPIKeyExchange=no", "-o", "ConnectTimeout=5"]
 
 
-if __name__ == "__main__":
-	send_lib.check_input_fields(sys.argv)
-	facility_name = sys.argv[1]
-	destination = sys.argv[2]
-	if len(sys.argv) == 3:
-		# if there is no destination type, use default 'host'
-		destination_type = send_lib.DESTINATION_TYPE_HOST
-	else:
-		destination_type = sys.argv[3]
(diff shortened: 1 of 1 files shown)


send/generic_sender.py
```diff
@@ -95,18 +95,20 @@ def prepare_ssh_transport_command():
...
-	destination = sys.argv[2]
-	if len(sys.argv) == 3:
-		# if there is no destination type, use default 'host'
-		destination_type = send_lib.DESTINATION_TYPE_HOST
-	else:
-		destination_type = sys.argv[3]
-		if destination_type == send_lib.DESTINATION_TYPE_EMAIL or destination_type == send_lib.DESTINATION_TYPE_SERVICE_SPECIFIC:
-			print("Destination type " + destination_type + " is not supported yet.", file=sys.stderr)
-			exit(1)
+def main(facility_name: str, destination: str, destination_type: str = 'host', opts=None) -> None:
```
xflord, line 98:
```suggestion
def main(facility_name: str, destination: str, destination_type: str = send_lib.DESTINATION_TYPE_HOST, opts=None) -> None:
```
nitpick

xflord (approved):
I like this. Sad that I didn't think of it.

send/generic_sender.py
zlamalp: Fixed, amended to previous commit.

HejdaJakub (approved):
It looks good from my side.

PR #827, opened 2023-09-15
feat(o365_mu_ukb_forward_status): add new service

kofzera:
- Added gen, send and slave scripts for new service.
- The service is a copy of o365_mu_account_status service.


xflord (changes requested):
Please also create the initial changelog entry

Johaney-s:
The scripts should be uploaded as executable files

Johaney-s (approved):

xflord (approved):

PR #831, opened 2023-09-29
feat(slack): service for propagation to Slack

Johaney-s:
* uses Slack SDK clients to communicate with Slack API
* each API (SCIM and WEB) have limited possibilities and our user can act as a user or a bot, based on the token
* bot has limited possibilities, but the main idea is that Perun manages those channels, where bot is integrated
* the propagation creates, updates and deactivates users, skips bots and guests
* also creates and archives channels and manages its members

DEPLOYMENT NOTE: Requires Slack SDK to be installed

### DIFF
--- a/gen/slack
+++ b/gen/slack
@@ -1,0 +1,90 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use perunServicesInit;
+use perunServicesUtils;
+use File::Basename;
+use JSON::XS;
+
+local $::SERVICE_NAME = basename($0);
+local $::PROTOCOL_VERSION = "3.0.0";
+my $SCRIPT_VERSION = "3.0.0";
+
+perunServicesInit::init;
+my $DIRECTORY = perunServicesInit::getDirectory;
+my $data = perunServicesInit::getHashedHierarchicalData;
+
+our $A_USER_LOGIN;                  *A_USER_LOGIN =                  \'urn:perun:user_facility:attribute-def:virt:login';
+our $A_RESOURCE_NAME;               *A_RESOURCE_NAME =               \'urn:perun:resource:attribute-def:core:name';
+our $A_RESOURCE_PRIVATE_CHANNEL;    *A_RESOURCE_PRIVATE_CHANNEL=     \'urn:perun:resource:attribute-def:def:isSlackPrivateChannel';
+our $A_RESOURCE_GENERAL_CHANNEL;    *A_RESOURCE_GENERAL_CHANNEL=     \'urn:perun:resource:attribute-def:def:isSlackGeneralChannel';
+our $A_MAIL;                        *A_MAIL =                        \'urn:perun:user:attribute-def:def:preferredMail';
+our $A_FIRST_NAME;                  *A_FIRST_NAME =                  \'urn:perun:user:attribute-def:core:firstName';
+our $A_LAST_NAME;                   *A_LAST_NAME =                   \'urn:perun:user:attribute-def:core:lastName';
+our $A_MEMBER_STATUS;               *A_MEMBER_STATUS =               \'urn:perun:member:attribute-def:core:status';
+
+my $struc = {"users", {}, "channels", {}};
+# if resource representing general channel is propagated, all propagated users must be included in it
+# if no such resource is propagated, all members of resources are taken as Slack users
+my $hasGeneralChannel = 0;
+
+foreach my $resourceId ($data->getResourceIds()) {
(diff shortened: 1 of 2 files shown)


send/slack
```diff
@@ -0,0 +1,518 @@
...
+
+
+def update_user(client, our_user, their_user):
+	print_debug(f"Updating user {our_user}")
+	user_data = User(
+		user_name=our_user["userName"],
+		name=UserName(
+			given_name=our_user["firstName"], family_name=our_user["lastName"]
+		),
+		emails=[UserEmail(value=our_user["email"])],
```
xflord, line 118: display_name is something we don't want to update?

send/slack
```diff
@@ -0,0 +1,518 @@
...
+			update_user(client, our_user, their_user)
+
+	return deactivated_users
+
+
+def create_channel(webclient, name, is_private):
+	print_debug(f"Creating {'public' if not is_private else 'private'} channel {name}")
+	try:
+		response = webclient.conversations_create(name=name, is_private=is_private)
+	except SlackApiError as e:
```
xflord, line 231: I don't think this ever gets raised (according to docs https://slack.dev/python-slack-sdk/api-docs/slack_sdk/#slack_sdk.WebClient.conversations_create). Also, why check for exception here and the `error` property in `invite_members`?

send/slack
```diff
@@ -0,0 +1,518 @@
...
+	response = webclient.conversations_unarchive(channel=channel_id)
+	check_webclient_response(response)
+	global CHANNELS_UPDATED
+	CHANNELS_UPDATED += 1
+
+
+def notify_users_deactivated(user_ids):
+	print_debug(f"Pushing notification about {len(user_ids)} deactivated users")
+	formatted_ids = ", ".join(["<@" + u + ">" for u in user_ids])
+	text = f"[UPOZORNĚNÍ] : Došlo k deaktivaci {'uživatelů' if len(user_ids) > 1 else 'uživatele'} {formatted_ids}.\n"
```
xflord, line 339: I suppose this should be in czech?

send/slack
```diff
@@ -0,0 +1,518 @@
...
+	response = webclient.chat_postMessage(text=text, channel=NOTIFICATIONS_CHANNEL_ID)
+	check_webclient_response(response)
+
+
+def should_kick(m, their_users):
```
xflord, line 344:
```suggestion
def should_kick(member, their_users):
```

send/slack
```diff
@@ -0,0 +1,518 @@
...
+
+	check_webclient_response(response)
+	global MEMBERS_ADDED
+	MEMBERS_ADDED += len(members)
+
+
+def kick_members(webclient, members, channel_id):
+	for user in members:
+		print_debug(f"Kicking user {user} from channel {channel_id}")
+		response = webclient.conversations_kick(channel=channel_id, user=user)
```
xflord, line 314: this could potentially be a lot of calls. Could this be a problem? (maybe the SDK does some buffering magic?) One SCIM API call should do the trick as well, no?

xflord (commented):
Best I can do

gen/slack
```diff
@@ -0,0 +1,90 @@
...
+our $A_MEMBER_STATUS;               *A_MEMBER_STATUS =               \'urn:perun:member:attribute-def:core:status';
+
+my $struc = {"users", {}, "channels", {}};
+# if resource representing general channel is propagated, all propagated users must be included in it
+# if no such resource is propagated, all members of resources are taken as Slack users
+my $hasGeneralChannel = 0;
+
+foreach my $resourceId ($data->getResourceIds()) {
+	my $isGeneralChannel = $data->getResourceAttributeValue(resource => $resourceId, attrName => $A_RESOURCE_GENERAL_CHANNEL );
+	if ($isGeneralChannel) {
```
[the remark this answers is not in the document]
Johaney-s: multiple resources representing one general channel are OK (for when there are members of more VOs involved - they need to have different resources)

send/slack
Johaney-s: i think the two approaches are the same, but i unified them.

send/slack
Johaney-s: for now it is sufficient. in case someone else would like to use it, we can make translations and set some language facility attribute

send/slack
Johaney-s: i didn't find bulk operation for that

xflord: Mixed up slack groups and channels, my bad.

xflord (approved):
code looks fine

PR #832, opened 2023-09-29
feat(hml_json): changed hml_json gen script to use scoped mu login

kofzera:


xflord (commented):
Please add deployment note that informs about change in required attributes

xflord (approved):

PR #833, opened 2023-09-29
fix(vsup_ifis): use only relations from stag

zlamalp:
- Transfer only STAG relations to IFIS.
- Remove all chars from VZTAH_ID before comparison.


xflord (approved):
changing the source db sounds like a feature more than a fix, but if it was broken beforehand makes sense

PR #840, opened 2023-10-10
docs: dhus documentation

mattjoke:
* documentation of the 'dhus' service
* added new structure to docs folder
* added files that links to other modules/files/concepts


docs/modules/generic_send.md
```diff
@@ -0,0 +1,77 @@
+# Generic Send (generic_send.py)
```
Johaney-s, line 1: IMHO would be better to describe generic_sender (which does pretty much the exact same, but should replace the generic_send in the future)

docs/modules/generic_send.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Description
+
+This script is used to send data generated by the [PerunDataGenerator](PerunDataGenerator.md) to the Perun API. It is
```
Johaney-s, line 5: the format of the generated data does not matter, it is not related to the general data generator at all

docs/modules/generic_send.md
```diff
@@ -0,0 +1,77 @@
...
+used by the service-specific sending scripts written in python.
```
Johaney-s, line 6: it can be called by send script written in any language, now most commonly by bash scripts. the service-specific wording can be misplaced with destination type service-specific, maybe it would be clearer to use some other word. it's just the send method that can be called directly by other python script.

docs/modules/generic_send.md
```diff
@@ -0,0 +1,77 @@
...
+First it checks the input parameters, and then it runs the complete generic sending logic.
+Based on the number of parameters sent to the script, it determines the destination type and calls the `send` method.
+
+### `send` method
+
+Sends data generated for the specified service/facility to the desired destination.
+Destination type can be specified as the last parameter, default is 'host'.
```
Johaney-s, line 13: opts is the last parameter

docs/modules/generic_send.md
```diff
@@ -0,0 +1,77 @@
...
+- 0 - success
+- &lt;other&gt; - call was successful, but there was an error on the destination side
+
+### `prepare_url_transport_command` method
+
+Prepares base transport command for CURL connection. If available, adds common perun certificate.
+
+Parameters:
+
+- `temp`: temporary file
```
Johaney-s, line 37: i feel like these parameters are not much interesting, I can after all see them in the code.
i would prefer a high-level description of how the sender works as a whole, what are the capabilities (destination types), what needs to be configured for which use case (credentials, credentials), what data can it transfer (any generated by gen script) etc.

docs/concepts/send.md
```diff
@@ -0,0 +1,20 @@
+# SEND script
+
+A bash script that transfers the generated data (as a .tar.gz archive) to the target machine.
```
Johaney-s, line 3: bash/perl/python, ... whichever

it needn't pass .tar.gz at all - it can call external API for example

docs/concepts/send.md
```diff
@@ -0,0 +1,20 @@
...
+
+The file name must match the service name.
+
+In a production environment, stored in `[path to engine]/send/`
+
+Each service script currently has to run a generic send script that takes care of everything. Eventually it will be
```
Johaney-s, line 9:
```suggestion
Some services currently run a generic send script that takes care of everything. Eventually it will be
```

docs/concepts/send.md
```diff
@@ -0,0 +1,20 @@
...
+possible to remove the send scripts for each service altogether. The current form of all send scripts is as follows:
```
Johaney-s, line 10:
```suggestion
possible to remove the send scripts for such services altogether. The current form of the simple send scripts is as follows:
```

docs/concepts/send.md
Johaney-s: can add link to the generic_send file

docs/services/webcentrum_eosc.md
```diff
@@ -0,0 +1,10 @@
+# webcentrum_eosc
```
Johaney-s, line 1: so this should be also separated to gen and send parts probably

docs/services/dhus.md
```diff
@@ -0,0 +1,21 @@
...
+
+### [SEND](../concepts/send.md)
+
+The script uses [generic_send.py](../modules/generic_send.md) to send data to the API.
+
+[//]: # (High-level description of the SLAVE part)
+
+### [SLAVE](../concepts/slave.md)
+
+Uses [perun-slave-base](../../slave/base) to call DESTINATION SCRIPT with the data.
```
Johaney-s, line 21:
```suggestion
Calls external script, which location needs to be set in a prescript. Also expects configuration file to be prepared on the target machine. The external script is then called with the parameters:
`script_name -f path_to_files -c path_to_configuration`
```

docs/concepts/pre-mid-post.md
```diff
@@ -0,0 +1,33 @@
...
+
+Runs `during` SLAVE script on the destination machine.
+A MID script is primarily used to make various modifications to the behavior of the script that differs from the
+standard
+use of the service by other service managers.
+
+## POST script
+
+Runs `after` SLAVE script on the destination machine.
+A POST scripts is primarily used for specific 'housekeeping', i.e. checking that everything is set as it should be,
```
Johaney-s, line 28:
```suggestion
A POST script is primarily used for specific 'housekeeping', i.e. checking that everything is set as it should be,
```

docs/concepts/gen.md
```diff
@@ -0,0 +1,15 @@
+# GEN script
```
Johaney-s, line 1: do we want to mention consents triggering here?

Johaney-s (commented):
some first thoughts, will think of it and report back again

xflord (commented):
Pretty much everything has been mentioned but:

- I would also prefer a higher-level practical perspective (let's assume our dear future co-worker can deduce parameters and such themselves)
- Separate folders for gen/send scripts would make sense to me, also documenting services that just call `generateMemberUsersDataInJson` seems redundant (It's clear right away to look at generic documentation)
- `Cieľom je aby dokumentácia high-level popisovala ako služba na technickej úrovni funguje` if we want to fulfill this I would include some little background about how/what data gets send to the scripts (which also includes mentioning consents).
-  A short `Creating new service` section might be good here, but might be redundant with what we have in Confluence

Otherwise really well organized, good job

Johaney-s:
@xflord I would prefer the documentation to be one file per service, I see the scripts as parts that cannot exist on their own, and it could be beneficial for the high-level view, for example if you want to write some general info about the service, it might not be clear to which file to put it.

docs/modules/generic_send.md
Johaney-s: maybe sections for each destination type would be suitable

xflord:
> @xflord I would prefer the documentation to be one file per service, I see the scripts as parts that cannot exist on their own, and it could be beneficial for the high-level view, for example if you want to write some general info about the service, it might not be clear to which file to put it.

As long as the file is clearly segmented (GEN, SEND, SLAVE) I'm happy. There are many times where I'm just trying to figure out how the send logic works, gen pretty much always just generates some data and passes it.

docs/concepts/gen.md
```diff
@@ -0,0 +1,20 @@
...
+
+Perl scripts that read and process data from Perun into data files for the services (or SLAVE scripts for
+further processing). These scripts are run by the Engine.
+
+The file name must match the service name.
```
xflord, line 6: This will not necessarily be true when we start using generic scripts (which is soon). I would change the wording a bit, the fact that gen/send script have to have the same file name seems more important to me

docs/concepts/send.md
```diff
@@ -0,0 +1,20 @@
+# SEND script
+
+A bash script that transfers the generated data to the target machine (e.g. by API call, tar, ssh, etc.).
```
xflord, line 3:
```suggestion
A script that transfers the generated data to the target machine (e.g. by API call, tar, ssh, etc.).
```
does not have to be bash

docs/concepts/send.md
```diff
@@ -0,0 +1,20 @@
...
+
+
+The file name must match the service name.
```
xflord, line 6: see above

docs/modules/generic_sender.md
```diff
@@ -0,0 +1,26 @@
+# Generic Send (generic_sender.py)
+
+## Description
+
+This script sends data (e.g. generated by the [PerunDataGenerator](PerunDataGenerator.md)) using API call, shh or tar command. It is
+used by the service-specific sending scripts (written in python/bash/perl etc.).
```
xflord, line 6: again as @Johaney-s mentioned before, `service-specific` can be misinterpreted

docs/concepts/gen.md
mattjoke: Yes, I did not know if I should write it as it is currently, or how it will be after generic JSON and support for multiple scripts.

docs/concepts/slave.md
```diff
@@ -0,0 +1,29 @@
...
+
+All script output (standard + error) is automatically returned to Perun.
+
+Available variables
+
+- All that you have defined in the PRE script.
+- `${WORK_DIR}` - reference to the service's work folder (contains all files as generated by the GEN script)
+- `${LIB_DIR}` - reference to the libraries folder, the full path to the helper file is ${LIB_DIR}/${SERVICE}/[filename]
+
+Available functions
```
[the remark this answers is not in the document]
xflord: `../../slave/base/bin/perun` should be the path I believe @mattjoke

docs/services/webcentrum_eosc.md
```diff
@@ -0,0 +1,21 @@
...
+
+### [GEN](../concepts/gen.md)
+
+The data is generated by
+the [generateMemberUsersDataInJson](../modules/PerunDataGenerator.md#generatememberusersdatainjson) method, not valid VO
+members are filtered out. The generated JSON data is stored in the` _webcentrum_eosc_ `file.
+
+
+[//]: # (High-level description of the SEND part)
```
sarkapalkovicova, line 10: Some forgotten comment?

sarkapalkovicova (commented):
Otherwise OK. Makes sense to me.

docs/concepts/pre-mid-post.md
```diff
@@ -0,0 +1,31 @@
+
+# PRE/MID/POST scripts
+The scripts are located in` /etc/perun/[service_name].d/`
+
+The expected filename is `[pre/mid/post]_[number]_filename.`
+
+There can be multiple scripts, they are run in alphabetical order, the `[number]` part, which typically starts with 10, is
+used to determine priority.
+
+#### Exmaples: `pre_10_set_variables.sh`, `mid_20_change_behavior.sh`, `post_30_housekeeping.sh`
```
HejdaJakub, line 10: If I understand it correctly, the order is always 1. pre, 2. mid and 3. post ... and the number is just for order scripts of the given type - `pre_10`, `pre_20`. This way the example is presented it could look like the number is the main order mechanism, but it should be (at least I think) just the second rule for ordering.

docs/README.md
```diff
@@ -0,0 +1,20 @@
+# Perun Services - Documentation
+
+This folder contains documentation for Perun Services. The documentation is written in Markdown and is generated
+using [MkDocs](https://www.mkdocs.org/).
+
+Table of concepts
+
+- [GEN](concepts/gen.md)
+- [SEND](concepts/send.md)
+- [SLAVE](concepts/slave.md)
```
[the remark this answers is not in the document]
mattjoke: Yes, I forgot to add those! Good catch!

docs/concepts/pre-mid-post.md
mattjoke: Yes, you are right. I did not understand it correctly at first, but you are right.

Johaney-s (approved):

kofzera (approved):

michalberky (approved):

sarkapalkovicova (approved):

PR #842, opened 2023-10-18
feat(generic_gen): send script for generic_gen

xflord:
* added send script for generic_gen
* the send script expects destination type every time (this should always get passed by the engine anyway)


xflord:
merge after https://github.com/CESNET/perun/pull/4077 and after gen script is implemented

send/generic_gen
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env python3
+import sys
+
+import generic_sender
+import send_lib
+
```
zlamalp, line 6: I would probably include some TODO or comment. It might be confusing to see "generic_gen" sending script, so we should explain, that this is due to the fact, that we have only single value for the script name/path, and we can't otherwise send data generated by the generic_gen script. I guess this should be also stated in the commit message.

xflord: Either way I was not sure whether the name would be `generic_send` so it is still more or less a place holder, but I added a comment + extended commit message

zlamalp:
Closed in source repository, closing here too.

PR #843, opened 2023-10-18
fix: 🐛 use LS ID in BBMRI_rt gen script

dBucik:
Switch from using BBMRI IDs and BBMRI usernames to usage of LS IDs and LS usernames

BREAKING CHANGE: 🧨 rt_bbmri now uses LS IDs and LS usernames

DEPLOYMENT NOTE: Service need to have `urn:perun:user:attribute-def:def:login -namespace:lifescienceid-persistent-shadow` and `urn:perun:user:attribute-def: def:login-namespace:lifescienceid-username` attributes assigned as required in Peurn configuration if they used BBMRI IDs or BBMRI usernames. See changes to get hint what attribute needs to be assigned for which service. Consents should be modified to be granted for these attributes, as the ID change is internal change and does not reall affect users.

### DIFF
--- a/gen/rt_bbmri
+++ b/gen/rt_bbmri
@@ -8,7 +8,7 @@ use JSON::XS;
 
 our $SERVICE_NAME = basename($0);
 our $PROTOCOL_VERSION = "3.1.1";
-my $SCRIPT_VERSION = "3.0.3";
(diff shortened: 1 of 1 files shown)


dBucik:
Switched to MR due to migration to GitLab. Still do not merge, please.

PR #848, opened 2023-10-30
feat(bbmri_negotiator): New service bbmri_negotiator

Johaney-s:


gen/bbmri_negotiator
```diff
@@ -0,0 +1,212 @@
...
+		# change from EXPIRED to VALID
+		$userStruc->{$uid}->{$u_status} = $status;
+	} elsif ($memberStatus eq $STATUS_SUSPENDED && $status eq $STATUS_VALID){
+		# change from SUSPENDED to VALID
+		$userStruc->{$uid}->{$u_status} = $status;
+	} elsif ($memberStatus eq $STATUS_SUSPENDED && $status eq $STATUS_EXPIRED){
+		# change from SUSPENDED to EXPIRED
+		$userStruc->{$uid}->{$u_status} = $status;
+	}
+	$userStruc->{$uid}->{$u_status} = $status;
```
xflord, line 191:
```suggestion
$userStruc->{$uid}->{$u_status} = $status;
```
Am I missing something? Maybe intended to be extended in the future?

gen/bbmri_negotiator
```diff
@@ -0,0 +1,212 @@
...
+	$userStruc->{$uid}->{$u_email} = $email;
+	$userStruc->{$uid}->{$u_name} = $d_name;
+	$userStruc->{$uid}->{$u_collections} = [];
+	$userStruc->{$uid}->{$u_biobanks} = [];
+	$userStruc->{$uid}->{$u_national_nodes} = [];
+	$userStruc->{$uid}->{$u_networks} = [];
+	$userStruc->{$uid}->{$u_adm_collections} = [];
+	$userStruc->{$uid}->{$u_adm_biobanks} = [];
+	$userStruc->{$uid}->{$u_adm_national_nodes} = [];
+	$userStruc->{$uid}->{$u_adm_networks} = [];
```
xflord, line 164: Not sure how this is supposed to work. You want to use strings as keys no? Ideally constants like $GROUP_ROLE_ADM_COLLECTION to reduce code in other methods as well

xflord (commented):
Not sure whether this was ready for review but here you go. Also please squash the commits when ready

gen/bbmri_negotiator
dBucik: Copy pasta from the bbmri_collections and bbmri_networks scripts. Not sure why this was implemented this way. But actually, we are only interested in the membership status in the BBMRI VO. It is also assumed that resources will be only for the BBMRI VO. So I think the whole mechanism can be removed (I am not even sure if the end service will care about the status).

Johaney-s: alright, we can then filter the non VALID members and not include the status in the result.

gen/bbmri_negotiator
```diff
@@ -0,0 +1,169 @@
...
+
+## creates users structure
+sub processGroups {
+	my $groupId = shift;
+	my $resourceId = shift;
+
+	my $roleIdValue;
+	my $groupRole;
+
+	if ($data->getGroupAttributeValue( group => $groupId, attrName => $A_G_COLLECTION_ID )) {
```
xflord, line 103: I'm assuming a group can only be one of these (biobank, network, etc.). In which case, wouldn't a check that only one of the attributes is set be nice? Or are the branches in this order because of "priority" of the attributes?

xflord (commented):
Other than that looks fine to me. You have an environment to test this on, right?

zlamalp:
This code has been merged in source repository. Closing this PR.

PR #852, opened 2023-11-06
feat(m365_cloud): allow change of group name

Johaney-s:
* match groups by perun uuid
* derive mailNickname from displayName

### DIFF
--- a/send/m365_cloud
+++ b/send/m365_cloud
@@ -7,6 +7,7 @@ import time
 import secrets
 import requests
 import send_lib
+import unicodedata
 
 # constants
 SERVICE_NAME = "m365_cloud"
@@ -54,6 +55,7 @@ users_created = 0
 users_updated = 0
 groups_created = 0
 groups_removed = 0
+memberships_updated = 0
 groups_updated = 0
 
 # language mapping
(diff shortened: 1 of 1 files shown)


send/m365_cloud
```diff
@@ -466,21 +483,21 @@ def remove_groups(our_groups, their_groups):
...
 	""" Checks which groups are missing in m365 and creates them, updates members of existing groups """
 	for our_group in our_groups:
-		their_group = next(filter(lambda g: g["displayName"] == our_group[groupname_attr_name], their_groups), None)
+		their_group = next(filter(lambda g: g.get(PERUN_ID_EXTENSION_NAME) == our_group[resource_id_attr_name], their_groups), None)
 		if their_group is None:
-			their_removed_group = next(filter(lambda g: g["displayName"] == our_group[groupname_attr_name], removed_groups), None)
-			if (their_removed_group and their_removed_group.get(PERUN_MANAGED_EXTENSION_NAME)
-				and their_removed_group.get(PERUN_ID_EXTENSION_NAME, -1) == our_group[resource_id_attr_name]):
+			their_removed_group = next(filter(lambda g: g.get(PERUN_MANAGED_EXTENSION_NAME, False)
+				and g.get(PERUN_ID_EXTENSION_NAME) == our_group[resource_id_attr_name], removed_groups), None)
```
xflord, line 497:
```suggestion
				and g.get(PERUN_ID_EXTENSION_NAME, -1) == our_group[resource_id_attr_name], removed_groups), None)
```
Not sure this is necessary but in case you missed it

xflord, line 497:
```suggestion
				and g.get(PERUN_ID_EXTENSION_NAME, -1) == our_group[resource_id_attr_name], removed_groups), None)
```
Not sure this is necessary but in case you missed it

xflord (approved):
Seems fine to me

send/m365_cloud
Johaney-s: when perun_managed attribute is true, the perun_id attribute should always be filled, so it is probably not needed. and uuid is always known, comparison with None would result to False.

PR #854, opened 2024-06-28
Add deleting of principals from kerberos

HavrilaJ:


zlamalp:
The code has been merged in source repository.

